PR #127, opened 2013-06-16
Removes original title directive from Add Group Input.

raghunayyar:
@tanghus @jancborchardt


tanghus:
Thanks for looking into this @raghunayyar 
It's not the optimal solution because somewhere the tipsy must be instantiated, but it solves the symptom :)

PR #128, opened 2013-06-16
Optimize pictures with optipng and scour

Niduroki:
Doing owncloud/core#2778 in here.

16 kB savings. Better than nothing I'd say.

@tanghus


tanghus:
thx

PR #185, opened 2013-09-02
Let user customize the size of images

Jonny007-MKD:
Hi,
I told my friend about OwnCloud and he complained about the bad resolution of the photos of his contacts after syncing them to his smartphone. This was something that I noticed before: My Samsung has a horizontal resolution of about 540px and when I did call somebody the full-width image looked quite screepy. So I decided to look into what is happening there and what could be done.

The result is this branch where the user is able to set the maximum height/width of the photos he is going to upload in the future. There is (afaik) no way to resize the existing images since they are not saved in their original resolution (perhaps the OwnCloud path is saved somewhere). As far as I comprehend [RFC 6350](http://tools.ietf.org/html/rfc6350#section-6.2.4) there is no limitation on image sizes.

Advantages:
- better images

Disadvantages:
- needs more disk space
- needs more time to sync (Setting the maximum size from 200px to 1000px enlarges the VCards from ~70kb to ~1MB). That's why I think it is best to let the user decide what size his images shall have.

ToDos:
- Decide what to do with existing images if the resolution is reduced: Shall they be resized? When?
- Add localization
- Review the changes

### DIFF
--- a/ajax/savecrop.php
+++ b/ajax/savecrop.php
@@ -29,6 +29,7 @@ namespace OCA\Contacts;
 
 require_once 'loghandler.php';
 
+$max_size = (int)\OCP\Config::getUserValue(\OCP\User::getUser(), 'contacts', 'max_size', 400);
 $image = null;
 
 $x1 = (isset($_POST['x1']) && $_POST['x1']) ? $_POST['x1'] : 0;
@@ -76,8 +77,8 @@ if($data) {
 			'savecrop.php, x: '.$x1.' y: '.$y1.' w: '.$w.' h: '.$h,
 			\OCP\Util::DEBUG);
 		if($image->crop($x1, $y1, $w, $h)) {
-			if(($image->width() <= 200 && $image->height() <= 200)
-				|| $image->resize(200)) {
+			if(($image->width() <= $max_size && $image->height() <= $max_size)
+				|| $image->resize($max_size)) {
 
 				// For vCard 3.0 the type must be e.g. JPEG or PNG
 				// For version 4.0 the full mimetype should be used.
--- a/ajax/uploadphoto.php
+++ b/ajax/uploadphoto.php
@@ -33,6 +33,7 @@ $l10n = OCA\Contacts\App::$l10n;
 $contactid = isset($_POST['contactid']) ? $_POST['contactid'] : '';
 $addressbookid = isset($_POST['addressbookid']) ? $_POST['addressbookid'] : '';
 $backend = isset($_POST['backend']) ? $_POST['backend'] : '';
+$max_size = (int)\OCP\Config::getUserValue(\OCP\User::getUser(), 'contacts', 'max_size', 200);
 
 if($contactid == '') {
 	bailOut('Missing contact id.');
@@ -50,8 +51,8 @@ if ($fn) {
 	$image = new OC_Image();
 	sleep(1); // Apparently it needs time to load the data.
 	if($image->loadFromData($data)) {
-		if($image->width() > 400 || $image->height() > 400) {
-			$image->resize(400); // Prettier resizing than with browser and saves bandwidth.
+		if($image->width() > $max_size || $image->height() > $max_size) {
+			$image->resize($max_size); // Prettier resizing than with browser and saves bandwidth.
 		}
 		if(!$image->fixOrientation()) { // No fatal error so we don't bail out.
 			debug('Couldn\'t save correct image orientation: '.$tmpkey);
@@ -98,8 +99,8 @@ if(file_exists($file['tmp_name'])) {
 	$tmpkey = 'contact-photo-'.md5(basename($file['tmp_name']));
 	$image = new OC_Image();
 	if($image->loadFromFile($file['tmp_name'])) {
-		if($image->width() > 400 || $image->height() > 400) {
-			$image->resize(400); // Prettier resizing than with browser and saves bandwidth.
+		if($image->width() > $max_size || $image->height() > $max_size) {
+			$image->resize($max_size); // Prettier resizing than with browser and saves bandwidth.
 		}
 		if(!$image->fixOrientation()) { // No fatal error so we don't bail out.
 			debug('Couldn\'t save correct image orientation: '.$tmpkey);
--- a/ajax/uploadphoto.php
+++ b/ajax/uploadphoto.php
@@ -33,7 +33,7 @@ $l10n = OCA\Contacts\App::$l10n;
 $contactid = isset($_POST['contactid']) ? $_POST['contactid'] : '';
 $addressbookid = isset($_POST['addressbookid']) ? $_POST['addressbookid'] : '';
 $backend = isset($_POST['backend']) ? $_POST['backend'] : '';
-$max_size = (int)\OCP\Config::getUserValue(\OCP\User::getUser(), 'contacts', 'max_size', 200);
+$max_size = (int)\OCP\Config::getUserValue(\OCP\User::getUser(), 'contacts', 'max_size', 400);
 
 if($contactid == '') {
 	bailOut('Missing contact id.');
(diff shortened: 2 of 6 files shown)


tanghus:
Thanks for the PR. I'll test it tomorrow.

js/config.php
```diff
@@ -36,3 +36,4 @@
 	. (OCP\Config::getUserValue($user, 'contacts', 'contacts_categories_indexed', 'no') === 'no'
 	? 'false' : 'true') . ',';
 echo 'lang=\'' . OCP\Config::getUserValue($user, 'core', 'lang', 'en') . '\';';
+echo 'max_size=\'' . OCP\Config::getUserValue($user, 'contacts', 'max_size', 'en') . '\';';
```
bartv2, line 39: default is wrong

ajax/savecrop.php
```diff
@@ -76,8 +77,8 @@
 			'savecrop.php, x: '.$x1.' y: '.$y1.' w: '.$w.' h: '.$h,
 			\OCP\Util::DEBUG);
 		if($image->crop($x1, $y1, $w, $h)) {
-			if(($image->width() <= 200 && $image->height() <= 200)
-				|| $image->resize(200)) {
+			if(($image->width() <= $max_size && $image->height() <= $max_size)
```
bartv2, line 80: @tanghus in other places the max_size was 400, is that important here?

Jonny007-MKD, line 80: I suppose this is due to the fact that the app only wanted to store
200px images, but it should also be possible for the user to crop the
images before they are resized to the end size so we will not always be
smaller than 200px.
So we have to consider if we want to double the value at the other
places, too, or do not resize at all since we cannot know how small the
part is the the user want to crop/cut out.

Am 03.09.2013 08:24, schrieb Bart Visscher:

tanghus, line 80: > I suppose this is due to the fact that the app only wanted to store 200px images

Exactly. I first resize the image to max 400px to have a reasonable size in the crop dialog. As mentioned in the comments elsewhere, GD also apparently does a better job of resizing than the browser, you transfer less data and you don't have to calculate the crop based on the original size.

> or do not resize at all since we cannot know how small the part is the the user want to crop/cut out.

The user could upload a 400x2000 px image, so we have to do some resizing.

ajax/uploadphoto.php
```diff
@@ -50,8 +51,8 @@
 	$image = new OC_Image();
 	sleep(1); // Apparently it needs time to load the data.
 	if($image->loadFromData($data)) {
-		if($image->width() > 400 || $image->height() > 400) {
-			$image->resize(400); // Prettier resizing than with browser and saves bandwidth.
+		if($image->width() > $max_size || $image->height() > $max_size) {
+			$image->resize($max_size); // Prettier resizing than with browser and saves bandwidth.
```
tanghus, line 55: I don't think we should make it configurable here for the reasons mentioned above. Then we would just get bug reports if people set their max_size to 2000 and the crop dialog doesn't fit in the browser ;)
Let's instead agree on a reasonable size for the initial image before cropping, and set that as an upper limit for max_size.

ajax/savecrop.php
Jonny007-MKD, line 80: Am 03.09.2013 14:22, schrieb Thomas Tanghus:

> In ajax/savecrop.php:
> 
> > @@ -76,8 +77,8 @@
> >             'savecrop.php, x: '.$x1.' y: '.$y1.' w: '.$w.' h: '.$h,
> >             \OCP\Util::DEBUG);
> >         if($image->crop($x1, $y1, $w, $h)) {
> > -           if(($image->width() <= 200 && $image->height() <= 200)
> > -               || $image->resize(200)) {
> > -           if(($image->width() <= $max_size && $image->height() <= $max_size)
> 
> ```
> I suppose this is due to the fact that the app only wanted to
> store 200px images
> ```
> 
> Exactly. I first resize the image to max 400px to have a reasonable
> size in the crop dialog. As mentioned in the comments elsewhere, GD
> also apparently does a better job of resizing than the browser, you
> transfer less data and you don't have to calculate the crop based on
> the original size.
> 
> ```
> or do not resize at all since we cannot know how small the part is
> the the user want to crop/cut out.
> ```
> 
> The user could upload a 400x2000 px image, so we have to do some resizing.
> 
> But the user could also upload a 4000x3000px group photo and only wants
> to use the head of one person which then would be about 300x200px. If we
> resize the photo to 400x300px first the resulting cropped image will be
> only 30x20px.

Do you think it is easy to let the browser send the maximum height &
maximum width it can display to so the user can view it full screen and
we save as much traffic as possible, and then use the resulting crop
rectangle and apply it on the original image (we would need some
calculation here)?

js/settings.js
```diff
@@ -202,4 +202,10 @@ $(document).ready(function() {
 			}
 		}
 	});
+
+	$('#max_size_input').on('change', function(event) {
+		event.preventDefault();
+		OC.Contacts.storage.setPreference('max_size', $(this).val());
+		$('#max_size_value').text($(this).val() + 'px');
+	});
 });
```
tanghus, line 211: settings.js isn't used anymore.

Jonny007-MKD, line 211: That is why I haven't found it in the template :D
Well, then one may put this code where it belongs.

ajax/savecrop.php
tanghus, line 80: I'd say no. This is a contact manager, not a photo manipulation app, so it should be as simple as possible. I'm OK with having a configurable max size with an upper limit, but personally I don't think the option should even be displayed in the settings - it is cluttered enough already.
Can we have a designers view on this @jancborchardt ?

js/settings.js
tanghus, line 211: > That is why I haven't found it in the template :D
> Well, then one may put this code where it belongs.

Yeah I still have some cleaning up to do before release.

ajax/savecrop.php
Jonny007-MKD, line 80: Yes, you're right about that. But we should also provide the correct
cropping functionality so the user does not end up doing it himself in
paint to get the correct resolution. And the user does not notice the
background processing that is done. Only the result is more like what he
expects.

I am also not happy with this option in the settings, because it is
nothing that needs to be configured every week and would need some
explaining text what it does. Perhaps the personal page fits more.

Personally I would be fine with a maximum width of ~600px because that
fits my phone, but every user's phone may have a different maximum
resolution (what about tablets?). I also don't care if the traffic is 10
times bigger. But what about other people in other countries that would
be happy  to decrease the size to 100px to save some bytes on their
mobile plans?
These are the reasons why I would make it configurable.

jancborchardt:
This should not be something to be decided by users. The best possible quality should be used, although a certain maximum is probably good. Something which makes the images look good even when in a normal size on retina.

tanghus:
Of relevance: #152 

We also have to remember:
- Phones show contact pics differently. The phones I've had only show a thumbnail.
- Giving the option to save larger pics puts a strain on both the phone and the server. I've had to optimize the contact loading heavily for people running on e.g. a Raspberry Pi, or the server would die.
- And again this is not a photo app ;)

The options I see where we could change the current behaviour are:
- Make the initial crop image a bit larger.
- Make the final resolution (where the image gets saved in the vcard) user configurable, but without having any UI settings for it.
- Make the cropper a resizable square as mentioned in #152 

Thanks for the effort @Jonny007-MKD but I'm gonna have to reject the PR in it's current form.

jancborchardt:
The arguably most used mobile operating systems, iOS and Android, both show big images for the person calling or being called. So that’s what we need to take into account. We don’t want people who are synchronized with ownCloud to look crappy. Also, we should probably show bigger thumbnails at least for frequently contacted people/favorites in the Contacts app UI in the future.

Granted, it would put strain on servers. Let’s just make it clear that »user configurable« in this case means »admin configurable«, if at all.

PR #226, opened 2013-10-17
Suppress sessions

Cymaphore:
Ability to supress session handling in dav requests, see https://github.com/owncloud/core/issues/5383 ... I would recommend to add this to the 5.x-fork as well.

### DIFF
--- a/carddav.php
+++ b/carddav.php
@@ -1,4 +1,5 @@
 <?php
+define("DISABLE_SESSION", 1);
 if(!file_exists('../../lib/base.php')) {
 	die('Please update the path to /lib/base.php in carddav.php or make use of /remote.php/carddav/');
 }


tanghus:
@DeepDiver1975 can you comment on this?

DeepDiver1975:
@tanghus @Cymaphore let's discuss this in the scope of OC7 again - we shall focus now on oc6 release.
Once we agreed on a solution in OC7 backporting can be discussed as well.

jancborchardt:
So let the discussion in the scope of ownCloud 7 begin.

tanghus:
@DeepDiver1975 @jancborchardt @Cymaphore we're in the scope ;)

in the scope

tanghus:
@Cymaphore Sorry but this has outlived itself. I appreciate taking up the issue. If it's still valid it would be great if you'd take it up again in core or on the mailing list.

PR #231, opened 2013-10-24
Fix app not loading any more in OC6 with PHP 5.3

PVince81:
The app doesn't load at all in OC6 Beta 1 with the following error:
"Using $this when not in object context at \/srv\/www\/htdocs\/owncloud\/apps\/contacts\/lib\/dispatcher.php#46"

This PR fixes that.

@tanghus please review.


tanghus:
Thanks @PVince81 
I've gotta make some proper integration tests. Currently Travis only checks a few classes.

PR #313, opened 2013-12-20
Add indexes for speedingup db queries

bartv2:
I noticed that there were no indexes defined, this should improve the db access a bit


tanghus:
Travis doesn't do any integration tests, but lets pull it into master and see how it goes :) It's normally PostgreSQL that barfs on indexes.
Thx.

PR #337, opened 2014-01-06
Paging contact list

ben-denham:
Fix for https://github.com/owncloud/contacts/issues/136 and https://github.com/owncloud/contacts/issues/334 that adds basic pagination for the contacts list.

When the contact list loads, it retrieves 15 contacts from each address book. Each time the new "load more" button is clicked, 15 more contacts will be retrieved from each address book, until there are no more contacts left to load.

While the paging limit defaults to 15 contacts per address book, it can be changed from the contacts settings.

Client side caching of contacts still operates, but cached contacts are now invalidated when the paging limit is changed as well as when the address book is modified.

While I see that pagination isn't under consideration because of other speed improvements to the contacts app, it was a feature that we required, and it seems that some others are interested in it too, so I thought that I would leave the changes here for consideration :)

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -56,6 +56,17 @@ $this->create('contacts_address_book', 'addressbook/{backend}/{addressBookId}')
 	)
 	->requirements(array('backend', 'addressBookId'));
 
+$this->create('contacts_address_book_page', 'addressbook/{backend}/{addressBookId}/{page}')
+	->get()
+	->action(
+		function($params) {
+			session_write_close();
+			$dispatcher = new Dispatcher($params);
+			$dispatcher->dispatch('AddressBookController', 'getAddressBook');
+		}
+	)
+	->requirements(array('backend', 'addressBookId'));
+
 $this->create('contacts_address_book_update', 'addressbook/{backend}/{addressBookId}')
 	->post()
 	->action(
--- a/css/contacts.css
+++ b/css/contacts.css
@@ -759,6 +759,22 @@ tbody tr.contact.active, tbody tr.contact:hover {
 	transition: all 0.5s ease-in-out;
 }
 
+#contact-load-more {
+    height: 3em;
+    width: 100%;
+    border-radius: 0;
+    border-width: 0 0 1px 0;
+    border-style: solid;
+    border-color: #DDD;
+    outline: none;
+    display: none;
+}
+
+#contact-load-more.loading {
+    text-indent: -3000px;
+    overflow: hidden;
+}
+
 @media screen and (min-width: 1400px) {
 	#contact > ul.propertylist {
 		-moz-column-count: 3;
--- a/css/contacts.css
+++ b/css/contacts.css
@@ -760,19 +760,19 @@ tbody tr.contact.active, tbody tr.contact:hover {
 }
 
 #contact-load-more {
-    height: 3em;
-    width: 100%;
-    border-radius: 0;
-    border-width: 0 0 1px 0;
-    border-style: solid;
-    border-color: #DDD;
-    outline: none;
-    display: none;
+	height: 3em;
+	width: 100%;
+	-moz-border-radius: 0; -webkit-border-radius: 0; border-radius: 0; 
+	border-width: 0 0 1px 0;
+	border-style: solid;
+	border-color: #DDD;
+	outline: none;
+	display: none;
 }
 
 #contact-load-more.loading {
-    text-indent: -3000px;
-    overflow: hidden;
+	text-indent: -3000px;
+	overflow: hidden;
 }
 
 @media screen and (min-width: 1400px) {
--- a/js/addressbooks.js
+++ b/js/addressbooks.js
@@ -9,6 +9,7 @@ OC.Contacts = OC.Contacts || {};
 		this.storage = storage;
 		this.book = book;
 		this.$template = template;
+	        this.nextPage = 1;
 	}
 
 	AddressBook.prototype.render = function() {
@@ -163,6 +164,11 @@ OC.Contacts = OC.Contacts || {};
 		return this.book.active;
 	};
 
+
+        AddressBook.prototype.getNextPage = function() {
+	    return this.nextPage;
+	}
+
 	/**
 	 * Save an address books active state to data store.
 	 * @param bool state
@@ -224,6 +230,10 @@ OC.Contacts = OC.Contacts || {};
 			$(document).trigger('status.contacts.error', response);
 		});
 	};
+
+    AddressBook.prototype.incrementNextPage = function() {
+	++this.nextPage;
+    }
 
 	/**
 	 * Controls access to address books
--- a/js/addressbooks.js
+++ b/js/addressbooks.js
@@ -9,7 +9,7 @@ OC.Contacts = OC.Contacts || {};
 		this.storage = storage;
 		this.book = book;
 		this.$template = template;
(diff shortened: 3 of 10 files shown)


jancborchardt:
Hm, there are some problems with this approach:

Instead of having a »Load more« button people need to click, the additional contacts should be loaded automatically when scrolling further down. Also, the scroll bar is so short that it’s confusing and seems as if you have very little contacts. It might give off the impression that not all your contacts are in there. It’s really cumbersome to get to contacts far down the list. Hence it’s a :-1: from me for now, sorry.

@ben-denham can you look into if you can solve it using something like a virtual list? See http://sergimansilla.com/blog/virtual-scrolling/

tanghus:
@ben-denham Thanks for the contribution. I can't test it now, so maybe you have already solved this, but have you tested the situation where a specific contact is requested on first load e.g. from the search function?
Like `http://localhost/owncloud/index.php/apps/contacts/#235`

ben-denham:
@tanghus I've tried out the search, and it only works if the user appears on the first page, but that should be easily solved with the adjustments I outline below.

@jancborchardt I've made a fork (https://github.com/ben-denham/virtual-list/tree/server-connected-virtual-list) of the virtual list that you posted, and have started working in a branch called **server-connected-virtual-list** to create a demo of how we could use virtual list that makes requests to the server for contacts as needed. If you clone the repo, go to the **server-connected-virtual-list branch**, and open /examples/index.html, you can try out the demo.

Points to note about the demo:
- Server requests for contacts are for a particular page (e.g. page 1 containing contacts 0-99). While it may be nicer to request contacts starting at a certain index (e.g. request 100 contacts starting with contact 123), this would not work with the current caching system based on etags.
- Server requests cannot be per address book, but will need to be for a single alphabetized list of all of a user's contacts, from every address book. This allows us to retrieve contacts for a certain fragment of the virtual list.
- If a different user adds or removes contacts while you are viewing the virtual list (in an address book that is shared between you), your virtual list will suffer no ill-effects when it loads new contacts (e.g. no missing or duplicated contacts on the next page), and it will remove potentially outdated, locally stored contacts. The refresh button also allows you to force the virtual-list to remove all locally stored contacts, and reload those that are currently being viewed. Currently in the contact app, all contacts are loaded at once, and I believe you only see someone else's changes if you refresh the page, so I think this would be an improvement.

For searching, we could make a request to the server to retrieve the list-index for a particular contact, and then scroll to that index in the virtual list.

Let me know if you think the approach I have taken in the demo and outlined here is acceptable, and I can start implementing the approach in this pull request.

Please also let me know if I haven't made anything clear or if you have any questions or suggestions for improvements :)

jancborchardt:
I don’t have time to test right now, but basically my point is: As long as this »paging« only works in the background and does not interfere with the interface (for example, people should not need to click »load more« buttons« or need to wait for additional contacts loading, etc.) it’s ok.

ben-denham:
@jancborchardt Yes, the only design change would be the addition of a refresh button to force the browser to re-retrieve the contacts from the server.

I think the question I have now is whether the way I have implemented the virtual list in my demo is viable to start applying to the contacts app, or whether anyone can see any flaws/points for improvement in the way I handle caching, loading contacts from the server, etc.

jancborchardt:
@ben-denham there shouldn’t be the need for a refresh button. Retrieving contacts should work in the background. Even pull-to-refresh is strange here because it’s just simple contacts. I don’t see a refresh button on Google Contacts or iCloud Contacts. ;)

@PVince81 @kabum @Raydiation maybe you’re interested looking at this virtual list approach as it might be useful for increasing performance of Files and other apps as well.

ben-denham:
@jancborchardt Ok, that's fine, we can remove the refresh button. Just to clarify though, could you please let me know what the expected functionality should be in this scenario:

I am viewing the contacts page, and I have scrolled through my entire list of contacts, so all of them are now stored locally in a JavaScript array. So no matter where I scroll to on the virtual list, I am no longer making requests to the server for new contacts. Now, say that I leave the contacts page open in my browser for several hours, and in that time, other users edit the contacts in my list (e.g. we share an addressbook that they can edit), therefore making the contacts in the list that I am viewing out of date.

Now, should the contacts in my list remain out of date until I refresh the page in my browser? Or perhaps the app should poll the server periodically to see if the local contacts are out of date (and if they are, clear the JavaScript array of contacts, meaning that the next time I scroll the list, the app will re-retrieve the contacts from the server). Or perhaps this should be handled in an entirely different way :)

Another quick note I'd like to add to this discussion, is that one extra complication I haven't taken into consideration in my demo is filtering the list by contact group, as obviously this would cause the indices of contacts in the virtual list would change.

Thanks for your help with refining this idea :)

MorrisJobke:
I would like to see it refreshed automatically. But we should discuss the behaviour when for example the contact that is currently open is remove by one of the "shared-to" persons. Will it just disappear?

jancborchardt:
@ben-denham periodic poll, yes. Whatever it takes to make sure automatic refresh happens as unobtrusively as possible in the background. :)

When a contact which is currently opened is removed by another person, maybe there should be a yellow notification up top saying »`contactname` has been deleted by `username`«. Then if you save the contact, it gets created again – if you close it, it vanishes.
But it seems like a thought different from the current pull request. :)

tanghus:
@ben-denham I just skimmed the code, and it looks very good. I've consciously avoided trying to implement this, because there are so many things to consider to make "lazy-loading" loading work properly, and - as others have also stated - I don't want paging like in Google contacts.

That being said I don't think the problem is in loading all contacts at a time, but rather rendering them. I would prefer keeping the contacts in memory, and instead invalidate DOM elements when not visible - maybe after a timeout?

The current caching using localStorage is very basic and invalidates the entire address book if one contact has changed. That could fairly easy be improved by fetching only changed contacts and update the cached data structure.

But to be honest neither of the approaches are anything I wanna be involved in coding. I'm trying to ease a bit down on the hours I spend on this project, and am very happy to see other people interested in contributing - especially on the front-end side :)

MorrisJobke:
@tanghus Maybe have a look at the links in https://github.com/owncloud/music/issues/78 (there are also non-angularJS ones ;))

tanghus:
@kabum I had https://github.com/sergi/virtual-list starred already :)

jbtbnl:
> That being said I don't think the problem is in loading all contacts at a time, but rather rendering them. I would prefer keeping the contacts in memory, and instead invalidate DOM elements when not visible - maybe after a timeout?

@tanghus check this out: https://github.com/owncloud/core/pull/7167

tanghus:
I had a quick glance at it yesterday, but there were so many changes in that PR, I couldn't find the actual scrolling part :smile: 
Besides that I have actually been convinced that moving to AngularJS would be the way to go, and plan beginning porting for oC7.

jancborchardt:
@jbtbnl @ben-denham what should we do with this PR then? It would probably be useful to have lazy loading in the contacts app, similar to Files and the user management.

@PVince81 can you maybe advise here regarding lazy loading?

jbtbnl:
@jancborchardt I think this PR does not fix the issue in the right way, instead we should do the same as in Files.

Can we agree to close this PR and create a new issue to start adressing this, plus https://github.com/owncloud/contacts/issues/576, https://github.com/owncloud/contacts/issues/595 and similar issues?

jancborchardt:
Yes, seems best. Closing this hence.

PR #340, opened 2014-01-07
Fix for #336

tsumi:
Incremented min-width in order to avoid glitch noticed in https://github.com/owncloud/contacts/issues/336


jancborchardt:
Tested by @schiesbn in the referred issue, seems to work fine for me as well. Hopefully will be superseded by the redesign from @jbtbnl anyway. :)

PR #355, opened 2014-01-12
Fix X-EVOLUTION-FILE-AS escape

DavidPrevot:
@smcv: Forwarding your proposal (updated for oC 6) via a pull request, (hopefully) full context provided bellow.

Evolution defaults to the form "X-EVOLUTION-FILE-AS:Bloggs\, Joe", and
since Evolution uses this field as the contacts' displayed name, it's
highly visible when ownCloud turns it into
"X-EVOLUTION-FILE-AS:Bloggs\, Joe".

https://github.com/owncloud/contacts/issues/42#issuecomment-28776546

When interacting with Evolution-Data-Server (which I think MeeGo might
use), the same incorrect escaping applies to X-EVOLUTION-FILE-AS. This
is a non-standard field, but it's highly visible for Evolution users,
since it's used as the default displayed name for contacts there.

https://github.com/owncloud/core/issues/2335#issuecomment-32130820

Bug-Debian: http://bugs.debian.org/735112


jbtbnl:
Great!

PR #388, opened 2014-01-30
remove %appswebroot% as it is completely useless

jancborchardt:
Please review @tanghus @jbtbnl @babelouest 

Next we need to change the icons which use %webroot% to use the newly defined CSS rules from core – but I’ll send a post to the mailing list about that.


jbtbnl:
Good one, plain CSS is always better.

tanghus:
Thanks :)

PR #401, opened 2014-02-03
Add ownCloud handle field

LEDfan:
For the Chat App we need an IM field so an owncloud user name to chat with can be added.

Is this okay? 
cc @tanghus @jancborchardt 

P.s. is there a way to retrieve the avatar of an contact?
Thanks

### DIFF
--- a/lib/utils/properties.php
+++ b/lib/utils/properties.php
@@ -132,6 +132,11 @@ Class Properties {
 					'xname' => 'X-SKYPE',
 					'protocol' => 'x-apple',
 				),
+				'owncloud_handle' => array(
+				    'displayname' => (string)$l10n->t('ownCloud handle'),
+				    'xname' => null,
+				    'protocol' => 'owncloud_handle'
+				),
 		);
 		if(is_null($im)) {
 			return $ims;
--- a/lib/utils/properties.php
+++ b/lib/utils/properties.php
@@ -135,7 +135,7 @@ Class Properties {
 				'owncloud-handle' => array(
 				    'displayname' => (string)$l10n->t('ownCloud handle'),
 				    'xname' => null,
-				    'protocol' => 'owncloud-handle'
+				    'protocol' => 'x-owncloud-handle'
 				),
 		);
 		if(is_null($im)) {
--- a/lib/utils/properties.php
+++ b/lib/utils/properties.php
@@ -132,10 +132,10 @@ Class Properties {
 					'xname' => 'X-SKYPE',
 					'protocol' => 'x-apple',
 				),
-				'owncloud_handle' => array(
+				'owncloud-handle' => array(
 				    'displayname' => (string)$l10n->t('ownCloud handle'),
 				    'xname' => null,
-				    'protocol' => 'owncloud_handle'
+				    'protocol' => 'owncloud-handle'
 				),
 		);
 		if(is_null($im)) {


jancborchardt:
Can you name it "owncloud-handle" with a dash please? Underscores are ugly and shouldn't be used here (the others like "x-apple" use dashes too).

tanghus:
Does the chat app use a special protocol?

> P.s. is there a way to retrieve the avatar of an contact?

Say you want the photo from a contact with id = 2, in your own address book with id = 1 you can call:

`http://owncloud/index.php/apps/contacts/addressbook/local/1/contact/2/photo`

LEDfan:
@jancborchardt Renamed it :) In the view it's new called "ownCloud" (ownCloud/chat#26)

@tanghus can it be merged?

> Say you want the photo from a contact with id = 2, in your own address book with id = 1 you can call:
> http://owncloud/index.php/apps/contacts/addressbook/local/1/contact/2/photo

Thanks.

tanghus:
I'd like to know the reason for needing this before merging.

jancborchardt:
Similarly as said in https://github.com/owncloud/chat/issues/26#issuecomment-34072360:

> The Contacts app should (much like Google Apps) automatically add people as contacts who are both 1) on the same ownCloud and 2) in a group where you can communicate with them. Like, when you can share a file with that ownCloud user, they should be displayed in your contacts.

Or in this case here: If I have a friend on my ownCloud and in my contacts, ownCloud should know the two entities are the same person. And if I have a friend on my ownCloud who is not in my contacts, they should appear in my contact list automatically.

LEDfan:
@tanghus yes indeed the Chat App is using it's own protocol, but will also support other protocols. The reason for the own protocol is in owncloud/chat#11

lib/utils/properties.php
```diff
@@ -132,6 +132,11 @@ public static function getIMOptions($im = null) {
 					'xname' => 'X-SKYPE',
 					'protocol' => 'x-apple',
 				),
+				'owncloud_handle' => array(
+				    'displayname' => (string)$l10n->t('ownCloud handle'),
+				    'xname' => null,
+				    'protocol' => 'owncloud_handle'
+				),
 		);
```
tanghus, line 140: Just to be a bit consistent I would suggest:

```
                'owncloud' => array(
                    'displayname' => (string)$l10n->t('ownCloud handle'), // Is handle needed?
                    'xname' => null,
                    'protocol' => 'x-owncloud'
                ),
```

tanghus:
> The reason for the own protocol is in owncloud/chat#11

Ehh, OK. I don't see anything related to that, but never mind.

Don't misunderstand me, I'm not at all against adding such functionality to contacts app; I just prefer it being done properly.

This approach is probably the simplest to get the functionality needed, and contact details can be found using OCP\Contacts.
But please change at least the protocol to `x-owncloud` or something. The `x-` part shows it as a non-standard protocol.

lib/utils/properties.php
jancborchardt, line 140: I would say yes, »handle« is needed. In the future with potential federation it would also include the address of the instance, possibly username@owncloud-instance.com (similar to Diaspora or Email).

And yeah, "x-owncloud" seems to be better.

LEDfan:
@tanghus updated your suggestion. 

Can it be merged now?

PR #423, opened 2014-02-25
Import

babelouest:
The import branch implements multiple file formats for importing addressbooks (vcard, csv, lidif).
It's based on the previous import module which was made for vcard import only, it should need a cleanup but it's ready for tests.


tanghus:
You need your branch rebase it to master first.

tanghus:
@babelouest Rebasing can be a PITA but needs to be done, so here goes ;

Update your master branch and then checkout your `import` branch.
From there run `git rebase master`.
You will get a message containing `CONFLICT (modify/delete): import.php deleted in HEAD and modified in import in progress.` and git stops.
Remove the obsolete `import.php` as that is not used anymore: `git rm import.php` and then as it says in the message run `git rebase --continue`.
You will then get a conflict in `lib/controller/importcontroller.php`. Fix that conflict and add it again with `git add lib/controller/importcontroller.php`.
Run `git rebase --continue` again. You will get more conflicts but continue this way and remember to either `git rm` or `git add` each conflict file before continuing each step.

Once all is fine commit the changes and make a new PR.

This really shows that it's important to often rebase your branch to avoid having too many conflicts ;)

tanghus:
Alternatively you can do a `git merge master` from your `import` branch. That _may_ be more painless ;)

tanghus:
btw https://help.github.com/ is really helpful :)

PR #432, opened 2014-03-03
MVC refactor, to leave only base class extensions in root app directory

alexisread:
I have had to label interface as interfaces, as Interface (and Abstract) is a PHP keyword. This is a slight deviation from the current naming convention, so a better name is welcome.

This is mainly prep work for mirroring the code structure here, in the files app, in order to extend the files app with categories.


tanghus:
@alexisread Looks like a good idea. It does look like you need to rebase it to current master first though.

alexisread:
Thanks, I've rebased the patch. I was having issues running the phpunit
tests to check though - I'm assuming that I just need to call eg. phpunit
addressbook_test.php?

tanghus:
> Thanks, I've rebased the patch. 

Oddly it still can't be merged automatically.

> I'm assuming that I just need to call eg. phpunit addressbook_test.php?

`phpunit --configuration tests/phpunit.xml --testsuite unit-tests`

tanghus:
@alexisread I'm closing this as it was never automatically mergable, and now master has changed even more. I will keep your suggestions in mind for a code restructuring post-oC7.

I'd appreciate it if you'd post an issue describing your thoughts behind this PR also in connection to the files app.

alexisread:
Hi, apologies for not finishing this off. I've updated it to the latest
trunk revision (which I believe is OC7), and I think I've managed to rebase
it correctly this time. My unit tests appear to be set up slightly
differently as I've had to alter the require path in bootstrap.php - I can
revert it if neccessary.

With regard to the files app changes, I'd like to move the files app into
the same MVC style as the contacts app. It'd be easier for me to then make
changes to the app ie. a sidebar and tags extension.

I had a question about the settings.php file in the contacts app root -
what is this entry point used for? The files app has several entry points
in root, most of which can be moved into controllers, but I wasn't sure
about settings.

Thanks,
Alexis

alexisread:
Hi, do you think you could find some time to look at my refactor? I'd like
to push this in to the code - should be up to date.
Thanks,
Alexis

PR #443, opened 2014-03-09
Import from multiple formats

tanghus:
A cleanup and squash of #426

PR #444, opened 2014-03-11
Import from multiple formats

tanghus:
Continuing from #443 _Sigh..._

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -155,7 +155,7 @@ $this->create('contacts_address_book_move_contact', 'addressbook/{backend}/{addr
 	)
 	->requirements(array('backend', 'addressBookId', 'contactId'));
 
-$this->create('contacts_import_upload', 'addressbook/{backend}/{addressBookId}/import/upload')
+$this->create('contacts_import_upload', 'addressbook/{backend}/{addressBookId}/{importType}/import/upload')
 	->post()
 	->action(
 		function($params) {
@@ -164,9 +164,9 @@ $this->create('contacts_import_upload', 'addressbook/{backend}/{addressBookId}/i
 			$dispatcher->dispatch('ImportController', 'upload');
 		}
 	)
-	->requirements(array('backend', 'addressBookId'));
+	->requirements(array('backend', 'addressBookId', 'importType'));
 
-$this->create('contacts_import_prepare', 'addressbook/{backend}/{addressBookId}/import/prepare')
+$this->create('contacts_import_prepare', 'addressbook/{backend}/{addressBookId}/{importType}/import/prepare')
 	->post()
 	->action(
 		function($params) {
@@ -175,9 +175,9 @@ $this->create('contacts_import_prepare', 'addressbook/{backend}/{addressBookId}/
 			$dispatcher->dispatch('ImportController', 'prepare');
 		}
 	)
-	->requirements(array('backend', 'addressBookId'));
+	->requirements(array('backend', 'addressBookId', 'importType'));
 
-$this->create('contacts_import_start', 'addressbook/{backend}/{addressBookId}/import/start')
+$this->create('contacts_import_start', 'addressbook/{backend}/{addressBookId}/{importType}/import/start')
 	->post()
 	->action(
 		function($params) {
@@ -186,9 +186,9 @@ $this->create('contacts_import_start', 'addressbook/{backend}/{addressBookId}/im
 			$dispatcher->dispatch('ImportController', 'start');
 		}
 	)
-	->requirements(array('backend', 'addressBookId'));
+	->requirements(array('backend', 'addressBookId', 'importType'));
 
-$this->create('contacts_import_status', 'addressbook/{backend}/{addressBookId}/import/status')
+$this->create('contacts_import_status', 'addressbook/{backend}/{addressBookId}/{importType}/import/status')
 	->get()
 	->action(
 		function($params) {
@@ -197,7 +197,7 @@ $this->create('contacts_import_status', 'addressbook/{backend}/{addressBookId}/i
 			$dispatcher->dispatch('ImportController', 'status');
 		}
 	)
-	->requirements(array('backend', 'addressBookId'));
+	->requirements(array('backend', 'addressBookId', 'importType'));
 
 $this->create('contacts_address_book_export', 'addressbook/{backend}/{addressBookId}/export')
 	->get()
--- a/formats/backend_ldap_connector.xml
+++ b/formats/backend_ldap_connector.xml
@@ -69,11 +69,6 @@
 			</not_null>
 		</ldap_core>
     
-    <ldif_entry name="displayname" enabled="true">
-      <vcard_entry property="FN">
-      </vcard_entry>
-    </ldif_entry>
-    
     <ldif_entry name="mail" enabled="true">
       <vcard_entry property="EMAIL" type="HOME">
       </vcard_entry>
--- a/formats/import_csv_gmail_connector.xml
+++ b/formats/import_csv_gmail_connector.xml
@@ -1,0 +1,229 @@
+<?xml version='1.0' standalone='yes'?>
+
+<!--
+	Parameters available
+	<import_core>
+	@brief: main parameters
+		<name></name>
+		@brief: Internal name
+		must match the xml file name with the pattern import_<name>_connector.xml
+		
+		<display_name></display_name>
+		@brief: Display name
+		
+		<type></type>
+		@brief: type (csv,vcard,ldif,...)
+		
+		<active></active>
+		@brief: if 0, will not be used
+		
+		<ignore_first_line><ignore_first_line/>
+		@brief: If the first line has titles (1 or true)
+		
+		<delimiter></delimiter>
+		@brief: if empty, automatic search between ',', ';' and '\t'
+		
+		<base_parsing></base_parsing>
+		@brief: parse by column name or column position
+		
+		<expected_columns></expected_columns>
+		@brief: if parse by column position, how many columns are expected (use for automatic import)
+	</import_core>
+
+	<import_entry name="" position="" enabled="" separator="" remove="">
+	@brief: csv entry to look for
+	name: property name (if base_parsing is name)
+	position: property position (if base_parsing is position)
+	enabled: import_entry used or not
+	separator: string used to separate multiple values
+	remove: string to remove in the value
+		
+		<altname lang=""></altname>
+		@brief: alternate property name
+		lang: language
+		value: other property name (if base_parsing is name)
+		
+		<vcard_favourites></vcard_favourites>
+		@brief: entry value to look for in order to add the current VCard to the Favourites section
+		
+		<vcard_entry property="" type="" prefix="" position="" group="" value="" separator="">
+		</vcard_entry>
+		@brief: VCard entry to add
+		property: VCard property name
+		type: VCard property type
+		prefix: prefix to add to the value
+		position: position of the element in the value
+		group: VCard group value
+		value: forced value
+		separator: separator string if position is set
+		
+		<vcard_parameter property="" parameter="">
+		</vcard_parameter>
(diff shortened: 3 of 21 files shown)


babelouest:
OK, found it, it was in your route url that was malformed line 288 :
`'addressbook/{backend}/{addressBookId}/{importType}/import/upload'` instead of `'apps/contacts/addressbook/{backend}/{addressBookId}/{importType}/import/upload'`.

But the import still doesn't work for me yet. The `ImportController.upload()` php function is called but not the following ones. Although upload() seems not to have any problem. May be a js issue.

tanghus:
> OK, found it, it was in your route url that was malformed line 288 :

Nice find!

>  but not the following ones.

They get called at my end, but I get the default app in the response again, although the route should be correct.

tanghus:
`{importType}` missing in the route for `Storage.prepare`, `Storage.start` and `Storage.importStatus`.

Two PRs conflict :P @DeepDiver1975 was kind enough to make a PR for the routing change, and I didn't notice the missing args in my 3 hour squashing seance :D

babelouest:
Arg, if I have looked at my emails in the last 15 minutes, I would have saved 15 minutes, we both found the solution, but you were first ! :p

It should work now with the last commit

tanghus:
\o/ I tested with the CSV i sent you, and the import worked very well besides the response from upload, which returns a count of 0: `{"filename":"AK-google.csv","count":0,"progresskey":"contacts-import-1202070064","backend":"local","addressBookId":"10"}`. It should be easily fixed as it's only a line count - 1.

VCF import with automatic detection gets the count right, but report `Imported {count} contacts` when done. Maybe because the last responses contained `{"progress":null}`?

BTW: With the Automatic format selected, the value is -1 which leads to a weird looking URL: `http://owncloud/index.php/apps/contacts/addressbook/local/7/-1/import/upload`. Shouldn't it shouldn't it just be `automatic`?

This is really great @babelouest !
Can you iron out those minor glitches so we can merge ASAP?

babelouest:
yeah, well I have looked at the code I have done and there are still efforts to made to make it clean.
in `templates/contacts.php`, the import_format select is filled in the same page, instead of a JQuery script.

I will change the automatic value from `-1` to `automatic` without too much problems I think.

Also, the progress information makes me ask more questions. Right now it's full of bugs because `importcontroller.php` is not really adapted to the multiple format import but is still looking for a VCard file, so the prepare function is parsing the file as a vcard to count elements.
In the current `importcontroller.php`, the number of cards is counted in upload(), prepare(), then start(). With the new import format, to count the number of cards, you have at least to parse the entire file with the function `OCA\Contacts\Connector\importConnector.getElementsFromInput()`.
My opinion is that this information (number of cards) is not relevant in the beginning or during the import process, and it should be given only at the end, then we could save tome cpu and developper time.

I would like to change the messages with the following:
- `starting file import`
- `format (selected|detected): {importFormat}`
- `processing {nb} card(s)` (this one will dynamically increase)
- `import end, {total} cards processed, {nbSuccess} successfull, {nbFail} errors`

What do you think ?

tanghus:
What about this

From an end user POV:
1. Select the file to import.
2. Once uploaded the format selector is shown with the (maybe) detected format preselected.
3. Press OK and import.

And the messages will be something like:
1. Server returns `{"importType": "vcard", "count": 123}`. Maybe add a `ImportManager::getCount($content, $type);` method? When you know the type counting is easy and it's an inexpensive call.
2. Client requests `/start` with the necessary arguments `{"filename":"somefile.vcf","progresskey":"contacts-import-1473403168", "importType": "vcard"}`. The user can have selected e.g. Yahoo vCard instead. 
3. Go.

> the import_format select is filled in the same page, instead of a JQuery script.

That's OK because the content is static. It's easy enough to change which option is selected, and maybe hide the irrelevant entries. (btw: I'd like to see a standard vCard entry too, so hiding others might be good)

tanghus:
@babelouest Another thing worth rethinking: 
The import process currently involves loading and parsing 7 XML before anything is done. The can be done a lot less CPU and IO intensive.
- For getting the correct file just match the importFormat with the filename.
- When detecting the format first do a match for file extension. 
  - If it's `.vcf` you match `import_vcard_*.xml` (to be sure read the first non-empty line and check for `BEGIN:VCARD`)
  - If it's csv you match for `import_csv_*.xml`, read the first non-empty line to match with.
  - For ldif there's only one.

Architecturally it may not be as elegant, but - without any stats - I guess you can cut initial processing to 1/10 or lower.

babelouest:
@tanghus , I like the idea of the import in 2 steps, lately, I have thought at this behaviour:
(the import type select is removed from the current UI)
- the user clicks the arrow button to upload a file
- a popup frame appears and shows a list of import types, highlighted with a color code corresponding to the calculated probability, in order of probability and with the first one selected by default
- the user validates the choice by clicking OK, then the file is imported

I don't think it's too heavy but someone told me to keep it simple, he didn't even want a dropdown list with import format at first, because he wanted it to be `automatic`. The current automatic procedure requires only one action from the user, 2 if the user selects manually the import format.
With your/my new behaviour, it would require 2 actions from the user all the time.
@jancborchardt , would you agree on this new behaviour or would you like to keep it 1 action for the automatic import ?

Concerning the XML parsing, I'm not sure it's that CPU or IO intensive, the class used is SimpleXML and I think that it has a pretty small memory print when used with small xml files. The xml files are quite small and used only in read mode.
I wouldn't like to add a specific extension verification because of that, I think it's too complicated and would cause further problems, like if the person has a `*.tmp` file and doesn't know about it, it would cause the import not to work.

I can make performance tests when the controller problems will be fixed, then we'll have a better idea if it's needed, do you agree ?

The standard VCard import type is easy, it's just a vcf connector file with no transformation, I'll make it when I'll have a few minutes.

tanghus:
> a popup frame appears and shows a list of import types

Instead of a popup let it replace the upload button, just like the current behaviour replaces import into/upload with the progress bar. It will be a short list of 1-4 choices, so maybe radio buttons will be best.

> highlighted with a color code corresponding to the calculated probability, in order of probability and with the first one selected by default. 

That could be icing on the cake :) To start with you could just order them by probability with the first/highest selected. If the probability is 100% (or how you calculate it), maybe simply start the import? That would please @jancborchardt ;)

> I can make performance tests when the controller problems will be fixed, then we'll have a better idea if it's needed, do you agree ?

Agreed.

babelouest:
```
It will be a short list of 1-4 choices,
```

I'm not sure putting more elements in this small area would be a good idea, I already find it too heavy for its size, that's why I came with the idea of a popup.
Anyway, I'm just starting to understand jquery, so either way, it will take a long time if I do it by myself :p

tanghus:
> I already find it too heavy for its size, that's why I came with the idea of a popup.

It is. I've been thinking of moving address books out of it and show them under Groups . People who only have one will not see much difference. People who do use more than one will generally want easier access to them.

But I'm digressing :)

I too will prefer a popup for dealing with the import. The UI it should be locked for interaction, because the list will update after the import. The feedback from users is also that it's too cramped.
Have a look at the merge dialog for an IMHO discreet - although modal - dialog.

> Anyway, I'm just starting to understand jquery, so either way, it will take a long time if I do it by myself :p

I can take care of that part :)

babelouest:
Here is a start, I cleaned a little bit, removed the cards count during the import process and display it only at the end, but for some reason, I still have this message at the end: `Processed {count} contacts`, it's because `self.storage.importStatus` is called without a {count} value in the parameters. I can't see why yet. If you have an idea.

tanghus:
Hmm. apparently github has changed so it doesn't show the code snippet when commenting inline, only a link to the line :(

babelouest:
I see your point. My explanation is the following.
Knowing the number of cards to import means reading the entire import file.

To prevent huge files import, I prefer doing that only in the `importController.start()` function, not in the `importController.upload()` or `importController.prepare()` functions.

Anyway, the progressbar evolves only during the `importController.start()` process so it won't be a problem to pass the total number to the progressbar, it's just that I don't understand the whole process with the php that talks to the js through the route and calls different php functions...

babelouest:
Sorry for the little mess.
I think we should agree on import messages and their order before changing them without notice (that's for me ;) )

What do you think of the one I suggested the other time :
- `starting file import`
- `format (selected|detected): {importFormat}`
- `processing {nb}/{total} card(s)` (this one will dynamically increase)
- `import end, {total} cards processed, {nbSuccess} successfull, {nbFail} errors`

tanghus:
> To prevent huge files import, I prefer doing that only in the importController.start() function, not in the importController.upload() or importController.prepare() functions.

Agreed.

> What do you think of the one I suggested the other time 

Sounds good. Then we don't have to maintain the total count client side.

>  it's just that I don't understand the whole process with the php that talks to the js through the route and calls different php functions...

The routing is actually like a DNS server (I've been messing with DNS the past days:P). It translates a route based on the URL and the HTTP method to a method in a controller.

tanghus:
@babelouest while your on it, can you take care of the issues mentioned here?

https://scrutinizer-ci.com/g/owncloud/contacts/inspections/338bb24b-9cd1-474f-927f-b3f38d92b4f8

babelouest:
@tanghus , ok, looks like cosmetic changes so it probably won't be bug material.

tanghus:
@babelouest Yes, it's "just" coding style, but nice to have fixed :)

babelouest:
I'm slowly moving on with what's left, althoug not all the scrutinizer changes are done yet, I think of https://scrutinizer-ci.com/g/owncloud/contacts/inspections/338bb24b-9cd1-474f-927f-b3f38d92b4f8/code-structure/operation/OCA%5CContacts%5CConnector%5CImportVCardConnector%3A%3AgetFormatMatch and https://scrutinizer-ci.com/g/owncloud/contacts/inspections/338bb24b-9cd1-474f-927f-b3f38d92b4f8/code-structure/operation/OCA%5CContacts%5CConnector%5CImportVCardConnector%3A%3AconvertElementToVCard but the other are OK I think.

I still have difficulties to understant the import on the js side though.
The `ImportController.getStatus()` function is called via the route to get the current status (obviously...) and sends the progresskey that it gots from the `OC::Cache`. The progresskey value seems to have only the number of processed cards.
Am I right at this point ?
I would like to use the getStatus() function to get more data like a string containing the current message `"importing {number} cards of {total}"`, and other response parameters containing the values `{number}` and `{total}` for example. Is it possible ?

I also don't understand the rest of the process where `storage.js` and `addressbook.js` are involved in importing and displaying the messages in the `#import-status-text` field.
If you have some clues to give me, that would be easier :p

tanghus:
Yes scrutinizer is very strict and I have made it even stricter so a build is marked as failed if there are ratings <= D :smile: 
I'm currently working through one of the refactoring suggestions for the PhotoController - it has many Ds...

I'm off for a couple of hours from now, but will go through the client<->server communication later tonight.

babelouest:
no problem, thanks !

tanghus:
Hey @babelouest 

I'm so tired I can hardly finish what I've started today, but tomorrow I've got the whole day, so I'll look at it then.

babelouest:
I did some changes, let me know if you're ok.
I could fix the import messages \o/

tanghus:
Great, looks good. Sorry I didn't get time to look at it yesterday. While refactoring I found a several bugs in AppFramework that needed to be fixed for contacts to work properly. Sucks maintaining the only app that uses core AppFramework :(
I'll give it a run down after breakfast :)

tanghus:
I'm only on phone for the weekend. If you have an XMPP account I can be reached at dukgo.com

babelouest:
Don't worry, I'm not in a hurry right now, it can wait 'till next week.
On my side, I don't see what more I can do in this PR, so there's no rush.

tanghus:
Feel free to push the merge button at will :)

babelouest:
nah, I'll wait for your go first.
I think we're close now but I'd like your feedback before sending it to the horde ;)

PR #457, opened 2014-04-01
OC users backend for the Chat app

LEDfan:
For the Chat app we need to have the ownCloud users in the contacts app. I wrote a backend and it works better than expected :)

There are still a lot todo's (implement error checking) I will fix them asap. But can you have already a look at it?

@tanghus I have a few questions about the permissions. It shouldn't be possible to delete this address book. Neither it should be possible to delete or create contacts. But it should be possible to change the personal information about users. How can this be achieved? I have played with \OCP\PERMISSION_\* but it didn't affected anything...

cc @jancborchardt


lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,316 @@
+<?php
+
```
LukasReschke: Please add the copyright notice at the top of the file.

LEDfan: Tnx done.

lib/backend/database.php
```diff
@@ -439,7 +440,6 @@ public function getContacts($addressBookId, array $options = array()) {
 			}
 
 		}
-
 		return $cards;
```
tanghus, line 443: I don't think this file should have been touched? ;)

LEDfan, line 443: Yes indeed. I edited some code in it for debugging. Should I undo it? Done.

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,336 @@
...
+	$query = \OCP\DB::prepare($sql);
+	$result = $query->execute(array($this->userid));
+	$row = $result->fetchRow();
+	$row['permissions'] = \OCP\PERMISSION_ALL;
+
+	return $row;
+    }
+
+    // Not needed since there is only one addressbook for every user
+    public function createAddressBook(array $properties) {
```
tanghus: Then don't implement it as the backend otherwise will think it's possible.

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,336 @@
...
+	    );
+	    $query = \OCP\DB::prepare($sql);
+	    $query->execute($args);
+
+	    return $this->getAddressBooksForUser();
+	} else {
+	    $row['permissions'] = \OCP\PERMISSION_ALL;
+	    return array($row);
+	}
+    }
```
tanghus: If there's only supposed to be one address book which won't change, there's no need to have a table for it. Just return and array containing hardcoded values. `displayname` and `description` should probably be translated.

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,336 @@
...
+	$sql = 'SELECT * FROM ' . $this->addressBooksTableName . ' WHERE id = ?';
+	$args = array($addressBookId);
+	$query = \OCP\DB::prepare($sql);
+	$result = $query->execute($args);
+	$row = $result->fetchRow();
+	$row['permissions'] = \OCP\PERMISSION_ALL;
+	$row['backend'] = $this->name;
+
+	return array($row);
+    }
```
tanghus: Again, this could be hardcoded. You could call this in `getAddressBooksForUser()` and return the result in an array.

tanghus:
> It shouldn't be possible to delete this address book. Neither it should be possible to delete or create contacts.

Then just don't implement those methods.

lib/backend/owncloudusers.php
LEDfan: Tnx.I will change it.

tanghus:
> But it should be possible to change the personal information about users. How can this be achieved?

In the metadata for `getContact()` and `getContacts()` set `permission` to `OCP\PERMISSION_UPDATE`

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,421 @@
...
+	    $query = \OCP\DB::prepare($sql);
+	    $result = $query->execute(array($this->userid));
+	    
+	    if (\OCP\DB::isError($result)) {
+		\OCP\Util::writeLog('contacts', __METHOD__. 'DB error: '
+		    . \OC_DB::getErrorMessage($result), \OCP\Util::ERROR);
+		return array();
+	    } else {
+		while($row = $result->fetchRow()){
+		    $row['permissions'] = \OCP\PERMISSION_ALL;
```
tanghus: This should be `\OCP\PERMISSION_UPDATE`

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,421 @@
...
+	    $query = \OCP\DB::prepare($sql);
+	    $result = $query->execute(array($this->userid));
+		    
+	    if (\OCP\DB::isError($result)) {
+		\OCP\Util::writeLog('contacts', __METHOD__. 'DB error: '
+		    . \OC_DB::getErrorMessage($result), \OCP\Util::ERROR);
+		return array();
+	    } else {
+		$row = $result->fetchRow();
+		$row['permissions'] = \OCP\PERMISSION_ALL;
```
tanghus: `\OCP\PERMISSION_UPDATE`

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,421 @@
...
+	    \OCP\Util::writeLog('contacts', __METHOD__.' exception: '
+		. $e->getMessage(), \OCP\Util::ERROR);
+	    return array();
+	}
+    }
+
+    // Not needed since there is only one addressbook for every user
+    public function createAddressBook(array $properties) {
+
+    }
```
tanghus: This should be removed or the backend thinks it can create address books

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,421 @@
...
+		if (is_null($id)) {
+		    \OCP\Util::writeLog('contacts', __METHOD__ . ' Couldn\'t find contact', \OCP\Util::ERROR);
+		    return false;
+		}
+
+	    } else {
+		throw new \Exception(
+		    __METHOD__ . ' If second argument is an array, either \'id\' or \'uri\' has to be set.'
+		);
+	    }
```
tanghus: The id/uri is only needed if the address book should be available via CardDAV. Don't know if that's the case?

tanghus:
Just to be sure I get it right: You keep a separate table so the user is able to add extra data to the users?

LEDfan:
For the question about the table: Yes.

lib/backend/owncloudusers.php
LEDfan: I don't think so.

tanghus:
BTW, I've set Scrutinizer to be pretty strict, so don't worry too much about that now :)

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,421 @@
...
+
+
+/**
+ * Contact backend for storing all the ownCloud users in this installation.
+ * Every user has *1* personal addressbook. The id of this addresbook is the 
+ * userid of the owner.
+ */
+class OwnCloudUsers extends AbstractBackend {
+
+    public $name = 'OwnCloudUsers';
```
tanghus: This should be the same as defined in app.php `owncloudusers`. Maybe it should just be `localusers`, `users` or something like that?

lib/backend/owncloudusers.php
LEDfan: Okay. But is it correct that the icons stay the same in the settings? So all the remove/update icons etc are present?

lib/backend/owncloudusers.php
LEDfan: `localusers` is indeed better

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,359 @@
...
+     * {@inheritdoc}
+     * Only 1 addressbook for every user
+     */
+    public function getAddressBook($addressBookId, array $options = array()) {
+	$addressbook = array(
+	    "id" => $addressBookId,
+	    "displayname" => 'ownCloudUsers',
+	    "description" => 'ownCloud Users',
+	    "ctag" => time(),
+	    "permissions" => \OCP\PERMISSION_ALL,
```
tanghus: Should be `\OCP\PERMISSION_READ`

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,359 @@
...
+		$contact = new Contact(	
+		    $addressBook = new AddressBook($this , $this->getAddressBooksForUser()), // since there is only one addressbook with OC users for each OC user we can use this function
+		    $this,
+		    array(
+			"id" => $user,
+			"lastmodified" => time(), 
+			"displayname" => \OCP\User::getDisplayName($user),
+			"fullname" => \OCP\User::getDisplayName($user)
+		    )
+		);
```
tanghus: This is a bit backwards. AddressBook and Contact are "models" which use the backend to for persistence.
Instead do something like:

``` php
$vcard = \Sabre\VObject\Component::create('VCARD');
$vcard->FN = \OCP\User::getDisplayName($user);
$now = new DateTime('now');
$vcard->REV = $now->format(\DateTime::W3C);
```

Then do the validation, set PRODID etc. Maybe we should have a helper method for that?

LEDfan: A helper function would be great :)

lib/backend/owncloudusers.php
LEDfan: Mm. I tried that, but then it's impossible to update contacts.

tanghus: Then it's an error in my code I guess :)

tanghus: Just checked. Yes, I take the permissions from the parent collection. AFAIRC I did that purposefully as that is how the CardDAV ACL works.
I'll have to improve that, but for now setting it to `\OCP\PERMISSION_UPDATE` _should_ work.

tanghus: abb3166e90a58df7ee0bb8e96208b1230f27de9d should fix it.

jancborchardt:
Nice! Will check it out soon. :) For now I assume that this adds a group called »ownCloud users« (or similar) to the groups list?

(I assume when all address books are enabled, contacts from all address books are displayed @tanghus?)

lib/backend/owncloudusers.php
LEDfan: Great this issue is solved :)
But the permissions of the contacts must be `\OCP\PERMISSION_ALL` otherwise I get a 403. I'm investigating it more.

tanghus:
> For now I assume that this adds a group called »ownCloud users« (or similar) to the groups list?

It will provide an "ownCloud users" address book. Groups doesn't have any backend. We can make it "look like" a groups though. The implementation details are less relevant.

> I assume when all address books are enabled, contacts from all address books are displayed?

That is the general idea, yes :wink:

lib/backend/owncloudusers.php
tanghus: > otherwise I get a 403. I'm investigating it more. 

When updating a contact?

tanghus: This may be because you instantiate a Contact object in the backend. You really shouldn't do that.

lib/backend/owncloudusers.php
```diff
@@ -0,0 +1,326 @@
...
+	    $query = \OCP\DB::prepare($sql);
+	    $result = $query->execute(array($this->userid));
+		    
+	    if (\OCP\DB::isError($result)) {
+		\OCP\Util::writeLog('contacts', __METHOD__. 'DB error: '
+		    . \OC_DB::getErrorMessage($result), \OCP\Util::ERROR);
+		return array();
+	    } else {
+		$row = $result->fetchRow();
+		$row['permissions'] = \OCP\PERMISSION_UPDATE;
```
LEDfan: I removed the initialization of the Contact class but, unfortunately it isn't working yet, I get an 304. If I change this to `\OCP\PERMISSION_READ` it works.
I'm trying to investigate this further.

LEDfan:
I removed the initialization of the Contact class but, and changed the permissions to `\OCP\PERMISSION_UPDATE`. But unfortunately it isn't working yet to load the contacts, I get an 304. If I change this to `\OCP\PERMISSION_READ` it works.
I'm trying to investigate this further.

tanghus:
What if you change it to `\OCP\PERMISSION_READ | \OCP\PERMISSION_UPDATE`?

LEDfan:
Hoera that works :)

tanghus:
:dancers: :cocktail:

jancborchardt:
> It will provide an "ownCloud users" address book. Groups doesn't have any backend. We can make it "look like" a groups though. The implementation details are less relevant.

@tanghus yeah, because all contacts are listed there then it would be cool if there automatically would be a group called »ownCloud users« to sort them under. Maybe even better: a section for »On this ownCloud« and under that all the groups you are part of. What do you think?

tanghus:
> Maybe even better: a section for »On this ownCloud« and under that all the groups you are part of. What do you think?

I think that gets too complex and is mixing user management with a contact manager. I see the use case, but imagine the app-navigation if the user has just a handful of groups, a couple of shared address books and X number of different user groups.

tanghus:
In another matter: We need either a global or per user or group setting on whether they want to appear in other users contacts. I have my personal ownCloud, which has a couple of accounts for friends and family. Neither of them are interested in showing up in other users contacts.
The issue is probably more important in a corporate or organizations level.
/cc @karlitschek

jancborchardt:
@tanghus yeah, that is probably confusing. But still can we do the first proposal, a group »On this ownCloud« for all the people in that »On this ownCloud« address book?

And another thing is: Some people on this ownCloud might be the same as contacts. That begs the question if a _separate_ address book is really the right way to go? Or rather adding »x-owncloud« attributes to each contact instead.

tanghus:
> But still can we do the first proposal, a group »On this ownCloud« for all the people in that »On this ownCloud« address book?

Absolutely

> And another thing is: Some people on this ownCloud might be the same as contacts. That begs the question if a separate address book is really the right way to go? Or rather adding »x-owncloud« attributes to each contact instead.

That's for now an implementation detail. We need a backend for knowing who is actually on this ownCloud instance. Let's wait with the discussion on how we can link/merge contacts 'till later.

LEDfan:
> > And another thing is: Some people on this ownCloud might be the same as contacts. That begs >> the question if a separate address book is really the right way to go? Or rather adding >>»x-owncloud« attributes to each contact instead.
> 
> That's for now an implementation detail. We need a backend for knowing who is actually on this >ownCloud instance. Let's wait with the discussion on how we can link/merge contacts 'till later.

A more important problem is that every owncloud user must be available as contact. If you let the users create contacts themselves by adding the x-owncloud-handle you can't be sure that when `foo` chat with `bar`, `bar` has the contact `foo`. And this is really necessary I hope I explain this clearly.

LEDfan:
BTW @tanghus for the Chat app I have to fetch _all_ contacts from the current user. I'm trying to use the `\OCP\ContactsManager::search()` function, which is mapped to the `addressbookprovider`. But I read that this is only for the local users? https://github.com/owncloud/contacts/blob/master/lib/addressbookprovider.php#L28
How should I fetch all the contacts? (especially the one from the backend I just created)

tanghus:
@LEDfan It should get registered here: https://github.com/owncloud/contacts/blob/master/appinfo/app.php#L62 but you're right that this needs porting.

I have some larger scale refactoring in mind that would solve this. But that also involves new interfaces in core, so for oC7 let's make a workaround and add a new backend method `getSearchProvider()` which returns a class instance that implements `\OCP\IAddressBook`. You would have to write that class too then.
Do you follow me?

Dang, there's still a lot of legacy code that needs rewriting :(

LEDfan:
Okay tnx. I'll try it.

LEDfan:
@tanghus just to be sure,  should I implement a fully search function, or just one which return all contacts?

jancborchardt:
> If you let the users create contacts themselves by adding the x-owncloud-handle you can't be sure that when foo chat with bar, bar has the contact foo. And this is really necessary I hope I explain this clearly.

@LEDfan right, good point. Then maybe it should be possible the other way around: from the »On this ownCloud« contacts to link them to other contacts you already have?

tanghus:
@LEDfan Write a class implementing `\OCP\IAddressBook` - I don't think it's worth it subclassing `AddressbookProvider` as it's messy and specialized.

And then add a method `getSearchProvider()` to `OCA\Contacts\Backend\LocalUsers` which returns an instance of that class. Perhaps it's an idea to inject `OCA\Contacts\Backend\LocalUsers` in the constructor if you can use some methods from that.

Something like:

``` php
public function getSearchProvider() {
    return new LocalUsersProvider($this);
}
```

I'll do the same for the other backends.

lib/backend/localusers.php
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     * Only 1 addressbook for every user
+     */
+    public function getAddressBook($addressBookId, array $options = array()) {
+	$addressbook = array(
+	    "id" => $addressBookId,
+	    "displayname" => 'Local Users',
```
jancborchardt: Can you change »Local Users« to »On this ownCloud«? »Users« is a term for drug addicts. ;)

tanghus: LOL

Wasn't there a decision not to use the brand name in texts btw?

jancborchardt: Yeah, not the name »ownCloud« in itself. But there’s a variable for it: `<?php p($theme->getName()); ?>`

LEDfan:
@tanghus thanks.

LEDfan:
@tanghus I just added the full search function. Integrating with the Chat app works :) (after some adjustments in `\OCP\Contactsmanager`) 

@tanghus is it possible to add the `\OCA\Contacts\Addressbook::getKey()` to every contact in `\OC\Contactsmanager`? (Like in  https://gist.github.com/LEDfan/9996155#file-contactsmanager-php-L46) I need this so the Chat client knows in which adressbook the avatar needs to bet fetched.

P.s. is it possible that there something with the temp avatar? If I try to access one via the url, I get the following error: `<b>Fatal error</b>:  Call to a member function fixOrientation() on a non-object in <b>/var/www/owncloud/apps/contacts/lib/utils/temporaryphoto.php</b> on line <b>186</b><br />`. Both in the Local and Localusers backend.

lib/backend/localusers.php
LEDfan: @tanghus @jancborchardt which should I use ? :P
- LocalUsers -> Local Users
- ownCloudUsers -> Owncloud Users
- ownCloudUsers -> On this p($them->getName())

jancborchardt: On this p($theme->getName())

:)

LEDfan:
@tanghus are there any things I should/can do before this can be merged?

tanghus:
@LEDfan Sorry, have been a bit under the weather. Will test it now. Hang on :)

lib/localusersaddressbookprovider.php
```diff
@@ -0,0 +1,163 @@
...
+	
+	while( $row = $result->fetchRow()) {
+	    $this->getProperty($results, $row);
+	}
+	return $results;
+    }
+    
+    public function getKey(){
+	
+    }
```
tanghus, line 93: You would have to to implement this...

lib/localusersaddressbookprovider.php
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+    * In comparison to getKey() this function returns a human readable (maybe translated) name
+    * @return mixed
+    */
+    public function getDisplayName(){
+
+}
```
tanghus, line 101: and this

lib/localusersaddressbookprovider.php
```diff
@@ -0,0 +1,163 @@
...
+    public function createOrUpdate($properties){
+	
+    }
+    
+    /**
+    * @return mixed
+    */
+    public function getPermissions(){
+	
+    }
```
tanghus, line 112: and this

But they can wait ;)

tanghus:
Seems to work very well.
- I have to fix not to show CardDAV URL and share links in UI.
- Have to fix that the delete button for users isn't shown.
- Is it OK that the ownCloud handle is deletable?

Any fixes can just be done in master

appinfo/database.xml
```diff
@@ -345,4 +345,130 @@
...
+			<name>addressbookid</name>
+			<type>integer</type>
+			<default></default>
+			<notnull>true</notnull>
+			<unsigned>true</unsigned>
+			<length>4</length>
+		</field>
+
+		<field>
+			<name>fullname</name>
```
tanghus, line 381: @LEDfan Change this to `displayname`. `fullname` is an artefact from like ownCloud 1 ;) I just don't want to change DB tables if I can prevent it as it can cause troubles.
If you watch the log you will notice an entry `Undefined index: displayname`

PR #459, opened 2014-04-03
WIP!!! Ldap

tanghus:
:construction: 

@babelouest I've made a PR for the branch to better be able to compare and comment on code


appinfo/routes.php
```diff
@@ -34,6 +34,17 @@ function($params) {
...
+$this->create('contacts_address_book_connectors', 'connectors/{backend}')
+	->get()
+	->action(
+		function($params) {
+			\OC::$session->close();
+			$dispatcher = new Dispatcher($params);
+			$dispatcher->dispatch('AddressBookController', 'getLdapConnectors');
+		}
+	)
+	->requirements(array('backend'));
```
tanghus, line 46: I think this should have a controller of its own

css/contacts.css
```diff
@@ -815,6 +815,27 @@ tbody tr.contact.active, tbody tr.contact:hover {
...
+#calendar-ui > p > textarea {
+	width:200px;
+	display: inline-block;
+}
+
+#calendar-ui > p > label {
+	vertical-align:top;
+	width:200px;
+	display: inline-block;
+}
```
tanghus, line 838: Please change all the references to calendar

js/addressbooks.js
```diff
@@ -67,7 +67,43 @@ OC.Contacts = OC.Contacts || {};
 			});
 		});
 		this.$li.find('a.action.edit').on('click keypress', function(event) {
-			if($(this).data('open')) {
```
tanghus, line 70: Make a check here:

``` javascript
if (!$(this).data('backend') === 'ldap') {
   // Open dialog
} else {
   // edit inline
}
```

js/addressbooks.js
```diff
@@ -650,3 +686,382 @@ OC.Contacts = OC.Contacts || {};
...
+					.fail(function(jqxhr, textStatus, error) {
+						var err = textStatus + ', ' + error;
+						console.log('Request Failed', + err);
+						defer.reject({error:true, message:error});
+					});
+				}
+			});
+		}
+	});
+}
```
tanghus: All the added code here should be moved to a separate object in a separate file.

js/jquery.ocaddnew.js
```diff
@@ -38,11 +38,32 @@
...
+/*					} else {
+						if(self.$ul.is('.open')) {
+							self.close();
+						} else {
+							self.open();
+						}
+					}*/
 					return false;
 				} else if($(event.target).is('button.primary')) {
 					var result = self.element.val().trim();
```
tanghus, line 48: As mentioned elsewhere: hands off this file ;)

js/loader.js
```diff
@@ -146,4 +146,6 @@ $(document).ready(function(){
 		FileActions.register('text/x-vcard','importaddressbook', OC.PERMISSION_READ, '', OC.ContactsImporter.init);
 		FileActions.setDefault('text/x-vcard','importaddressbook');
 	}
+	
+	$("#calendar-ui").hide();
```
tanghus: Huh?

js/storage.js
```diff
@@ -127,7 +127,7 @@ OC.Contacts = OC.Contacts || {};
 		return this.requestRoute(
 			'addressbook/{backend}/add',
 			'POST',
-			{backend: 'local'},
+			{backend: backend},
```
tanghus, line 130: Thanks. That had gone unnoticed for a while :)

js/storage.js
```diff
@@ -196,6 +196,32 @@ OC.Contacts = OC.Contacts || {};
...
+	 * {
+	 * metadata:
+	 * {
+	 *     id:'1234'
+	 *     permissions:31,
+	 *     displayname:'My contacts',
+	 *     lastmodified: (unix timestamp),
+	 *     owner: 'joye',
+	 * }
+	 */
```
tanghus, line 216: Please update the docs

lib/addressbook.php
```diff
@@ -181,9 +182,10 @@ public function getChild($id) {
 		}
 
 		// When requesting a single contact we preparse it
-		if (isset($this->objects[$id])) {
-			$this->objects[$id]->retrieve();
-			return $this->objects[$id];
+		if (isset($curContact)) {
+			$curContact->retrieve();
+			$this->objects[$id] = $curContact;
+			return $curContact;
```
tanghus, line 188: I don't see the reason for this change?

lib/backend/abstractbackend.php
```diff
@@ -328,8 +328,8 @@ public function setModifiedAddressBook($addressBookId) {
 	 * @param VCard $contact
 	 * @param array $options - Optional options
 	 * @return string|bool The identifier for the new contact or false on error.
-	public function createContact($addressBookId, $contact, array $options = array());
 	 */
+	public abstract function createContact($addressbookid, $contact, array $options = array());
```
tanghus: No no no, don't change these. They are only there for documentation. The backend checks if these methods are implemented. If they are it indicates that the backend can create/update/delete. If you change this you can't create a read-only address book.
Please in PRs keep to the subject of the PR.

lib/backend/abstractbackend.php
```diff
@@ -447,14 +447,53 @@ public function getPreferences($addressBookId) {
...
 		$key = 'prefs_' . $key;
+		
+		$addressbookList = $this->getAddressbookList($params);
+		/*if (!is_array($addressbookList)) {
+			$addressbookList = array($addressbookList);
+		}*/
+		if (!in_array($addressbookid, $addressbookList)) {
+			$addressbookList[] = $addressbookid;
+			$this->setAddressbookList($addressbookList);
+		}
```
tanghus: Here you are making a generic method specific. Instead add the addressbookList to the `$params` array before calling `setPreferences()`

lib/backend/abstractbackend.php
```diff
@@ -447,14 +447,53 @@ public function getPreferences($addressBookId) {
...
+			? \OCP\Config::setUserValue($this->userid, 'contacts', $key, $data)
+			: false;
+	}
+	
+	protected function getAddressbookList() {
+		$key = $this->name . "_list";
+		$data = \OCP\Config::getUserValue($this->userid, 'contacts', $key, false);
+		
+		return $data ? json_decode($data) : array();
+	}
```
tanghus, line 466: Move the last two methods to the LDAP backend.

lib/backend/ldap.php
```diff
@@ -357,7 +355,74 @@ public function updateAddressBook($addressbookid, array $properties, array $opti
...
+		}
+	}
+
+	/*
+	 * Sets the parameters for a new or existing addressbook
+	 *
+	 * @param array $properties
+	 * @return string|false The ID if the newly created AddressBook or false on error.
+	 */
+	public function setAddressBook(array $properties, array $options = array()) {
```
tanghus, line 372: Is this called from anywhere else than this backend?

lib/backend/ldap.php
```diff
@@ -385,7 +451,7 @@ public function deleteAddressBook($addressbookid, array $options = array()) {
 	 * @returns int | null
 	 */
 	public function lastModifiedAddressBook($addressbookid, array $options = array()) {
-    return null;
+		return null;
```
tanghus: No need to implement this. It's already implemented in AbstractBackend.

lib/controller/addressbookcontroller.php
```diff
@@ -175,7 +174,8 @@ public function updateAddressBook() {
 		$response = new JSONResponse();
 
 		$addressBook = $this->app->getAddressBook($params['backend'], $params['addressBookId']);
-		$addressBook->update($this->request['properties']);
+		//$addressBook->update($this->request['properties']);
+		$addressBook->update($this->request->post);
```
tanghus, line 178: What was the reason for this change?

appinfo/routes.php
babelouest, line 46: So far I only need this one, do you think it's better to create a controller just for this ?

css/contacts.css
babelouest, line 838: Can I name it `#file` or `#bookmark` then ?
just kidding, `#addressbook` it is ;)

js/addressbooks.js
babelouest, line 70: OK but there is a bug with the group add that shares the same event, how can I split them ?

Also, the dialog window can create and edit local addressbooks

js/jquery.ocaddnew.js
babelouest, line 48: sorry, didn't know :)

js/loader.js
babelouest: I know, I know, better use the css for that

lib/addressbook.php
babelouest, line 188: I mentionned that a few days ago, https://github.com/owncloud/contacts/issues/412#issuecomment-38934377 , it's a mystery but when I was working on it, the original version didn't work, while this one did, I don't understand why...

lib/backend/abstractbackend.php
babelouest: Sorry again, sometimes I get too proactive...

lib/backend/abstractbackend.php
babelouest: But then a function like that should be added

``` php
abstract function getAddressbookList() {
```

Maybe it's better to move all those in the ldap backend

lib/backend/abstractbackend.php
babelouest, line 466: So I must abstract them in the abstractbackend, otherwise there will be conflicts, or another solution ?

lib/backend/ldap.php
babelouest, line 372: in the `createAddressbook` and `updateAddressbook` methods

lib/backend/ldap.php
babelouest: https://github.com/owncloud/contacts/blob/ldap/lib/backend/abstractbackend.php#L263
it's empty ?

lib/controller/addressbookcontroller.php
babelouest, line 178: I'll get back to you on that one :)

appinfo/routes.php
tanghus, line 46: In the long run it's better to keep it separated. You might need other routes for LDAP, so a controller for that is the best.

css/contacts.css
tanghus, line 838: He he

js/addressbooks.js
tanghus, line 70: > OK but there is a bug with the group add that shares the same event, how can I split them?

What do you mean? I'm guessing it's because you messed with `jquery.ocaddnew.js`. Don't do that ;)

js/loader.js
tanghus: I just meant what it was doing in that file? The loader is to trigger an import dialog from the files app.

lib/addressbook.php
tanghus, line 188: Forgot about that. It is indeed weird 'cause it should be exactly the same..?

lib/backend/abstractbackend.php
tanghus: ROFL

lib/backend/abstractbackend.php
tanghus: See below ;)

lib/backend/abstractbackend.php
tanghus, line 466: No just move them to the LDAP backend. They're only used there.

lib/backend/abstractbackend.php
```diff
@@ -447,14 +447,53 @@ public function getPreferences($addressBookId) {
...
 			? \OCP\Config::setUserValue($this->userid, 'contacts', $key, $data)
 			: false;
 	}
+	
+	public function removePreferences($addressbookid) {
+		$key = $this->combinedKey($addressbookid);
+		$key = 'prefs_' . $key;
+		
+		\OC_Preferences::deleteKey( $this->userid, 'contacts', $key );
+
```
tanghus, line 465: This is a good addition to the abstract backend.

lib/backend/abstractbackend.php
```diff
@@ -447,14 +447,53 @@ public function getPreferences($addressBookId) {
...
+		
+		\OC_Preferences::deleteKey( $this->userid, 'contacts', $key );
+		
+		$addressbookList = $this->getAddressbookList($params);
+		$toRemove = array_search($addressbookid, $addressbookList);
+		if (is_int($toRemove)) {
+			unset($addressbookList[$toRemove]);
+			$addressbookList = array_values($addressbookList);
+			$this->setAddressbookList($addressbookList);
+		}
```
tanghus: But move the rest here to the LDAP backend

lib/backend/ldap.php
```diff
@@ -301,17 +289,19 @@ public function getAddressBook($addressbookid, array $options = array()) {
 			return $this->addressbooks[$addressbookid];
 		}
 		// Hmm, not found. Lets query the db.
-		$preferences = self::getPreferences($addressbookid);
+		$preferences = (array)self::getPreferences($addressbookid);
```
tanghus: You probably shouldn't cast to array if it can also be false.

lib/backend/ldap.php
tanghus, line 372: Yes but only in this backend? It's no biggie, but I prefer to make internal methods protected, so future maintainers - and myself - don't make any hacky shortcuts ;)

A better name may be `setAddressBookProperties()` btw?

lib/backend/ldap.php
tanghus: Yes, empty methods returns null.

appinfo/routes.php
babelouest, line 46: Since I made it generic by using the backend as parameter, what about creating a controller called `backendController` ?

tanghus, line 46: The currently only method is very specific, so why no call it LdapController? You will probably also need routes for updating the settings.

babelouest, line 46: Not that I'm obsessed with a more generic controller but if in a near future we implement a `Gmail Addressbook` backend or an `Exchange` backend for example, I'd rather mutualize the efforts.

tanghus, line 46: You've got a point there.

js/addressbooks.js
babelouest, line 70: No, I mean the dialog window is also usable for local addressbooks, it shows only `name` input when you select a local backend

js/addressbooks.js
babelouest: I'm not very familiar with the object structure, here is my code that doesn't work:

``` javascript
OC.Contacts = OC.Contacts || {};


(function(window, $, OC) {
    'use strict';

    var Dialog = function(storage, addressbook) {
        this.storage = storage;
        this.addressbook = addressbook;
    }

    Dialog.prototype.openAddressbookUi = function() {
```

and in addressbook.js, I use 

``` javascript
var dialog = OC.Contacts.Dialog(null, null);
```

which doesn't exists...

Can you help me with the structure of the new object ?

js/addressbooks.js
tanghus, line 70: Ah, thaddaway. I just think it's overkill to open a dialog to edit one field.

js/addressbooks.js
tanghus: Have you added it to a separate file? In that case you need to load it. Scripts and style sheets are added here: https://github.com/owncloud/contacts/blob/master/lib/controller/pagecontroller.php#L54

babelouest: I have moved it to a separate file, now it's the new object structure I'm not familiar with...

tanghus: Can you push it to this branch? Will make it easier to comment on.

Side note: I hope you don't think it's a PITA that I just comment and say "what's wrong or right". I'm by no means an expert on this shit - look at all the crappy code I've made - but doing it like this makes it a learning experience for both of us.

babelouest: > I hope you don't think it's a PITA that I just comment and say "what's wrong or right". 

Not at all, you're very helpful, I don't take all your feedbacks as a "you're pitiful" comment, on the contrary.
As I said, it's my first experience with JQuery and owncloud UI programming, so it's normal that I make mistakes and you help me correct them.
Please continue ;)

tanghus: > I hope you don't think it's a PITA that I just comment and say "what's wrong
> or right".
> 
> Not at all, you're very helpful, I don't take all your feedbacks as a
> "you're pitiful" comment, on the contrary.

Glad to hear that. It's certainly not meant that way :)

I've kinda lost track of code comments by now, and it seems github has so as 
well, so I suggest we make a fresh PR soon to continue the discussion on.

I'll probably be wasted most of the weekend, so no need to hurry ;)

PR #460, opened 2014-04-04
Work In Progress, Ldap Backend, second try

babelouest:
From the last PR, I remember the following major improvements that were not already done :
- create a class to host the `dialog.js`  functions
- update the documentation
- use ocdialog instead of JQuery dialog

@tanghus , feel free to add more ;)


babelouest:
@tanghus , a js class has been created for the dialog ui, next steps are using ocdialog and improving the documentation.

babelouest:
@tanghus , eventually, I was able to switch to ocdialog: https://github.com/owncloud/contacts/commit/e354d0fae37a5abf69201267ef122dd92b917ec3
The other problems seem to be fixed too.
One new problem linked to ocdialog appeared, ocdialog doesn't automatically resize when elements appear and disappear, do you know how to fix that ?

Nevertheless, can you go around the new dialog window and give me some feedbacks ?

templates/contacts.php
```diff
@@ -507,3 +509,121 @@ class="coords"
 	</span>
 </li>
 </script>
+
+<form id="addressbooks-ui" method="GET" name="addressbook-ui" action="#">
```
tanghus: You have to "hide" the template or it will be evaluated by the browser. Wrap it in a fake script tag:

``` php
<script id="addressBookConfigTemplate" class="hidden" type="text/template">
<div id="addressbooks-ui-div" title="Add new addressbook" class="addressbooks-ui-class">
...
</div>
```

The `type="text/template"` will prevent it from being evaluated by the JS engine.

js/dialog.js
```diff
@@ -0,0 +1,361 @@
...
+							console.log('Request Failed', + err);
+							defer.reject({error:true, message:error});
+						});
+					}
+				});
+			}
+		});
+	};
+  
+})(window, jQuery, OC);
```
tanghus: Since this is very specialized for LDAP I would prefer another name for the file instead of the generic `dialog.js`. `ldapconfig.js`?

js/dialog.js
```diff
@@ -0,0 +1,361 @@
+OC.Contacts = OC.Contacts || {};
+
+(function(window, $, OC) {
+	'use strict';
+
+	var AddressBookDialog = function() {
```
tanghus: I'd also prefer a more descriptive name instead of `AddressBookDialog`. Maybe `LdapConfig`?

tanghus: Here in the constructor you save the references to the storage and template:

``` javascript
var AddressBookDialog = function(storage, $template) {
    this.storage = storage;
    this.$template = $template;
};
```

I prefix jQuery objects with `$` so that you know later on you don't have to jQuery wrap the again with e.g. $(template).

js/dialog.js
```diff
@@ -0,0 +1,361 @@
...
+
+	AddressBookDialog.prototype.openAddressbookUi = function() {
+		var self = this;
+		$('#addressbooks-ui')[0].reset();
+		$('#addressbooks-ui-uri').attr('disabled', false);
+		$('#addressbooks-ui-ldapanonymous').attr('checked', false);
+		$('#addressbooks-ui-ldapreadonly').attr('checked', false);
+		$('#addressbooks-ui-ldapuser').attr('disabled', false);
+		$('#addressbooks-ui-ldappass').attr('disabled', false);
+		$('#addressbooks-ui-ldapbasednmodify').attr('disabled', false);
```
tanghus: To use boolean values you'd have to use `.prop('disabled', true)`

js/dialog.js
```diff
@@ -0,0 +1,361 @@
...
+
+		$('#addressbooks-ui-backend').change(function() {
+			var storage = OC.Contacts.storage;
+			self.addressbookUiInit();
+			$.when(storage.getConnectors($('#addressbooks-ui-backend').val()))
+			.then(function(response) {
+				$('#addressbooks-ui-ldapvcardconnector').empty();
+				var $option = null;
+				for (var id in response.data) {
```
tanghus: Looping javascript arrays has some pitfalls: http://stackoverflow.com/a/3010848/373007
Better use `for (var i = 0; i < response.data.length; i++) {`

js/dialog.js
```diff
@@ -0,0 +1,361 @@
...
+					if (response.data[id] != null) {
+						$option = $('<option value="' + response.data[id].id + '">' + response.data[id].name + '</option>');
+						$('#addressbooks-ui-ldapvcardconnector').append($option);
+					}
+				}
+				$option = $('<option value="">' + 'Custom connector' + '</option>');
```
tanghus: Translate?

js/dialog.js
```diff
@@ -0,0 +1,361 @@
...
+				var err = textStatus + ', ' + error;
+				console.log('Request Failed', + err);
+				defer.reject({error:true, message:error});
+			});
+		});
+		this.addressbookUiInit();
+	};
+
+	AddressBookDialog.prototype.editAddressbookUI = function(addressbook) {
+		var storage = OC.Contacts.storage;
```
tanghus: When passed in the constructor you've already got it.

templates/contacts.php
```diff
@@ -507,3 +509,121 @@ class="coords"
...
+<div id="addressbooks-ui-div" title="Add new addressbook" class="addressbooks-ui-class">
+	<input type="hidden" id="addressbooks-ui-addressbookid" />
+	<p id="addressbooks-ui-backend-p">
+	<label for="addressbooks-ui-backend">
+	<?php p($l->t('Addressbook type')); ?>:
+	</label>
+	<select id="addressbooks-ui-backend">
+		<option value="local">Local</option>
+		<option value="ldap">LDAP</option>
+	</select>
```
tanghus: We shouldn't be able to change to the type. Let's now over-complicate it.

tanghus:
I'm sorry @babelouest but I'm not very comfortable with the current UI approach. It's way too complex and for 99% of the uses it doesn't add value, only complexity. Can you split out the UI parts for later, and have it so that LDAP address books can be edited by hand?
We can add them after we've properly discussed how to implement it. I think it's better in a new PR, and I'd like to have @jancborchardt and @jbtbnl in for design discussions on that.

templates/contacts.php
babelouest: OK, done

js/dialog.js
babelouest: OK, renamed `otherbackendconfig.js`

babelouest:
@tanghus , that's ok, I can remove the backend selection in the dialog window to keep just the ldap for now.
I'd add a second link `"Add ldap addressbook"` below the first one and when the user clicks on the edit link, if it's a local use the old modification input, otherwise use the ldap dialog if it's a ldap addressbook.
That would be less overkill for a local addressbook.

js/dialog.js
babelouest: done

js/dialog.js
babelouest: heeeeee translate what ?
this is where the connector select is filled, through the `backendController.getConnector()` function.

js/dialog.js
babelouest: Now I do :)

templates/contacts.php
babelouest: This select was disabled when editing an existing addressbook, but now it's not there anymore.

babelouest:
@tanghus , new commit: https://github.com/owncloud/contacts/commit/c674b8db00faca842bc8b99b07c803cfceb08f98
is supposed to implement/fix your feedbacks

babelouest:
@tanghus , scrutinizer seems very maniac, useful but maniac, he finds duplicated code where it's not... :)

babelouest:
@tanghus , I have the following error message when remove an addressbook (local or ldap):
`PHP Fatal error:  Call to a member function getUserId() on a non-object in contacts/lib/controller/addressbookcontroller.php on line 213`
I don't recall changing this, would it be from you ?

tanghus:
@babelouest Can you merge master again? I just pushed some updates

tanghus:
Yeah, sometimes Scrutinizer has some false positives ;)

tanghus:
Sorry for not following much these past days. Feeling a bit under the weather and only pushing fixes to keep up with core.

babelouest:
That's ok, no problem.
If you have time, you can check the ldap backend result by using this public ldap server connection params:
x500-ldap-params

babelouest:
@tanghus , concerning the delete bug (which is also in master), I have made a patch: add 2 simple lines in `lib/controller.php` to instanciate `$this->api`

``` php
// In the use statement
\OC\AppFramework\Core\API

// in __construct
$this->api = new API('contacts');
```

But I wouldn't create a branch to make a PR, is there a simpler way to make a PR without creating a branch ?

tanghus:
>  concerning the delete bug (which is also in master), 

Sorry about that. A result of trying to keep up with core :P Fixed in the commit referenced above.

tanghus:
Totally unrelated, but I forgot where in Canada you're located? I have different TZs on my desktop clock to know when I have a chance of reaching people, and yours had got lost in an update :)

babelouest:
I live in Quebec, so my TZ is EST, UTC - 5 hours (same as NY, Toronto, Boston, etc.) :)

babelouest:
@tanghus , I was wondering why the group and favorite management was not working with the ldap backend, then I saw this: https://github.com/owncloud/contacts/blob/ldap/lib/controller/groupcontroller.php#L229
And realized the groups/addto route doesn't send the backend or addressbookid in the GET or the POST parameters :(
I'll open an issue.

Also, I'm wondering how the favorite flag works. Is it a CATEGORY or another property value or is it stored elsewhere ?

tanghus:
As mentioned in the other issue groups are only for local, own contacts. And yes, favourites is a "special" group :)

babelouest:
@tanghus , since it seems that beta testers are hard to find, and I would like to merge this branch to master, how about adding a checkbox in the administrator section, like:
`Enable LDAP backend (Warning, BETA)` which would enable/disable the link `add ldap addressbook`, then merge it ?

tanghus:
@babelouest Sounds like a good plan. Looks like you'll have to merge master once again first though :)

babelouest:
@tanghus , here it is: https://github.com/owncloud/contacts/commit/a3166ee45232d325a554402a60a94d9d00614564
It also can be reused for future backends

babelouest:
@tanghus , I'm having problems while rebasing the code, I have the following error message:
`Call to a member function getKey() on a non-object in owncloud/core/lib/private/contactsmanager.php on line 102`

In the `contactsmanager.php` file, the following code:

``` php
$this->address_books[$address_book->getKey()] = $address_book;
```

throws an error since `$address_book` seems null, do you have an idea why ?

tanghus:
Have you updated core?

templates/contacts.php
```diff
@@ -23,6 +23,15 @@
 					<ul class="addressbooklist">
 					</ul>
 					<input type="text" tabindex="0" autofocus id="add-address-book" placeholder="<?php p($l->t('Display name')); ?>" title="<?php p($l->t('Add Address Book')); ?>" />
+					<?php
+					if (\OCP\Config::getAppValue('contacts', 'backend_'.$backend, "false")) {
```
tanghus: Where does this `$backend` come from?

babelouest: That's a very good question, it comes from nowhere since it doesn't exist :p
fixed

babelouest:
yeah, I did, didn't do the trick though, but then I tried to rebase this branch, didn't go very well...
I'll retry in the next days

babelouest:
@tanghus last problem was due to `LdapBackend->getSearchProvider()` that didn't exist in the first place. I reused the `database`'s one, seems to work like that.

Rebased apparently successfully now. If you have no objection, I'd merge now

tanghus:
@babelouest Yeah sorry, my bad. I added an empty implementation to the abstract class, and had intended to make measures for preventing errors, but had some personal problems so forgot all about it.

If it works on your end go on and merge ASAP.

Please add an issue with the configuration for a public (or other?) LDAP server that we can use for testing. Or drop me a mail with it, but I'd prefer if as many as possible can test. I'm a bit repellent to trying to set up a server of my own :P

Then lets get the import UI sorted out and go into intensive test-mode :)

PR #465, opened 2014-04-16
Register localusers addressbook in app.php

LEDfan:
@tanghus I have added the registering of the localusers addressbook to the appinfo/app.php

### DIFF
--- a/lib/app.php
+++ b/lib/app.php
@@ -56,7 +56,7 @@ class App {
 		'ldap' => 'OCA\Contacts\Backend\Ldap',
 		'local' => 'OCA\Contacts\Backend\Database',
 		'shared' => 'OCA\Contacts\Backend\Shared',
-		'localusers' => 'OCA\Contacts\Backend\LocalUsers',
+		//'localusers' => 'OCA\Contacts\Backend\LocalUsers',
 	);
 
 	public function __construct(
@@ -152,7 +152,7 @@ class App {
 		if (!$info) {
 			throw new \Exception(self::$l10n->t('Address book not found'), 404);
 		}
-		
+
 		$addressBook = new AddressBook($backend, $info);
 		self::$addressBooks[] = $addressBook;
 		return $addressBook;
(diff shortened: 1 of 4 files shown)


lib/app.php
```diff
@@ -56,7 +56,7 @@ class App {
 		'ldap' => 'OCA\Contacts\Backend\Ldap',
 		'local' => 'OCA\Contacts\Backend\Database',
 		'shared' => 'OCA\Contacts\Backend\Shared',
-		'localusers' => 'OCA\Contacts\Backend\LocalUsers',
+		//'localusers' => 'OCA\Contacts\Backend\LocalUsers',
```
LEDfan, line 59: @tanghus is it okay/necessary to delete it here?

tanghus:
@LEDfan I'm afraid we misunderstood each other.

What I meant was to add a method `getSearchProvider()` to `\OCA\Contacts\AddressBook` which returns `$this->getBackend()->getSearchProvider($this);`

Then in `appinfo/app.php` change to:

``` php
    $app = new App($api->getUserId());
    $addressBooks = $app->getAddressBooksForUser();
    foreach ($addressBooks as $addressBook)  {
        if ($addressBook->isActive()) {
            \OCP\Contacts::registerAddressBook($addressBook->getSearchProvider());
        }
    }

```

LEDfan:
@tanghus oh yes indeed. I will update the code.

tanghus:
Maybe check for a result first? Not sure if all backends will implement a searchprovider. Currently only local, shared and localusers are relevant.

LEDfan:
Okay, and then if there isn't one use `\OCA\Contacts\Addressbookprovider` right?

tanghus:
Hmm, no then don't register it.

tanghus:
I've added an empty `getSearchProvider()` method to `AbstractBackend` so it won't barf if a backend hasn't implemented it.

LEDfan:
Replaced by #470 (so I can start from master again)

PR #467, opened 2014-04-16
add space between on this and Name

vgezer:
No strings will need to be changed on transifex.

### DIFF
--- a/lib/backend/localusers.php
+++ b/lib/backend/localusers.php
@@ -98,8 +98,8 @@ class LocalUsers extends AbstractBackend {
     public function getAddressBook($addressBookId, array $options = array()) {
 	$addressbook = array(
 	    "id" => $addressBookId,
-	    "displayname" => self::$l10n->t('On this') . self::$defaults->getName(),
-	    "description" => self::$l10n->t('On this') . self::$defaults->getName(),
+	    "displayname" => (string)self::$l10n->t('On this %s', array(self::$defaults->getName())),
+	    "description" => (string)self::$l10n->t('On this %s', array(self::$defaults->getName())),
 	    "ctag" => time(),
 	    "permissions" => \OCP\PERMISSION_READ,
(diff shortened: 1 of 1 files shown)


lib/backend/localusers.php
```diff
@@ -98,8 +98,8 @@ public function getAddressBooksForUser(array $options = array()) {
     public function getAddressBook($addressBookId, array $options = array()) {
 	$addressbook = array(
 	    "id" => $addressBookId,
-	    "displayname" => self::$l10n->t('On this') . self::$defaults->getName(),
-	    "description" => self::$l10n->t('On this') . self::$defaults->getName(),
+	    "displayname" => self::$l10n->t('On this') . ' ' . self::$defaults->getName(),
+	    "description" => self::$l10n->t('On this') . ' ' . self::$defaults->getName(),
```
tanghus: It should probably be

`"displayname" => self::$l10n->t('On this %s',  array(self::$defaults->getName())`

@LEDfan ?

LEDfan: Yes I think so...

vgezer: @tanghus @LEDfan Nope. This gives only {displayName} instead. I tried that already :/

tanghus: Really? Weird. Have you tried 
`"displayname" => self::$l10n->t('On this %s', array((string)self::$defaults->getName())` ?
It's no biggie if not possible, but interpolating variables is preferred.

tanghus: Ah, now I know:

``` php
"displayname" => (string)self::$l10n->t('On this %s', array(self::$defaults->getName()),
```

Translation strings are objects that needs to be cast to string. That happens automatically if you concatenate a string to them.

vgezer: Thanks. Now pushed the changes again.

tanghus:
Thanks :)

PR #481, opened 2014-05-05
Use  instead of

LEDfan:
@tanghus This is needed to solve errors on PHP 5.3'. (I know I shouldn't use php5.3 but it's default on centos.....)

### DIFF
--- a/lib/dispatcher.php
+++ b/lib/dispatcher.php
@@ -80,7 +80,7 @@ class Dispatcher extends MainApp {
 		});
 		$this->container->registerService('GroupController', function(IAppContainer $container) use($appName, $app) {
 			$request = $container->query('Request');
-			$tags = $container->server->getTagManager()->load('contact');
+			$tags = $container->getServer()->getTagManager()->load('contact');
 			return new GroupController($appName, $request, $app, $tags);
 		});
 		$this->container->registerService('ContactController', function(IAppContainer $container) use($appName, $app) {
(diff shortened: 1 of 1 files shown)


LukasReschke:
Hehe - PHP 5.3 is still supported by ownCloud so don't feel ashamed ;-)
A good IDE can help finding such things while developing - try to get a PHPStorm license ;-)

lib/dispatcher.php
```diff
@@ -80,7 +80,7 @@ public function registerServices() {
 		});
 		$this->container->registerService('GroupController', function(IAppContainer $container) use($appName, $app) {
 			$request = $container->query('Request');
-			$tags = $this->server->getTagManager()->load('contact');
+			$tags = $container->server->getTagManager()->load('contact');
```
tanghus: This should be

``` php
$tags = $container->getServer()->getTagManager()->load('contact');
```

lib/dispatcher.php
```diff
@@ -89,7 +89,7 @@ public function registerServices() {
 		});
 		$this->container->registerService('ContactPhotoController', function(IAppContainer $container) use($appName, $app) {
 			$request = $container->query('Request');
-			$cache = $this->server->getCache();
+			$cache = $container->server->getCache();
```
tanghus: Likewise here

lib/dispatcher.php
```diff
@@ -98,7 +98,7 @@ public function registerServices() {
 		});
 		$this->container->registerService('ImportController', function(IAppContainer $container) use($appName, $app) {
 			$request = $container->query('Request');
-			$cache = $this->server->getCache();
+			$cache = $container->server->getCache();
```
tanghus: And here

tanghus:
Thanks for catching these issues @LEDfan 
Can I get you to correct the errors mentioned before I merge?

LEDfan:
@tanghus the errors are fixed.

tanghus:
That was fast :)

PR #482, opened 2014-05-05
Update DisplayName when changed refs #474

LEDfan:
cc @tanghus

### DIFF
--- a/lib/backend/localusers.php
+++ b/lib/backend/localusers.php
@@ -376,7 +376,7 @@ class LocalUsers extends AbstractBackend {
 	}
 
 	public function updateDatabase($addressBookId) {
-		$sql = 'SELECT * FROM ' . $this->cardsTableName . ' WHERE addressbookid = ?';
+		$sql = 'SELECT id, displayname FROM ' . $this->cardsTableName . ' WHERE addressbookid = ?';
 		$query = \OCP\DB::prepare($sql);
 		$result = $query->execute(array($addressBookId));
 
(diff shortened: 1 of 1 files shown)


lib/backend/localusers.php
```diff
@@ -403,10 +405,33 @@ public function updateDatabase($addressBookId) {
...
+			$this->updateDisplayNames($contacts);
 			return true;
 		}
 	}
-
+	
+	private function updateDisplayNames($contacts){
+		foreach($contacts as $id => $displayname){
+			if(\OCP\User::getDisplayName($id) !== $displayname){
+				$this->updateDisplayName($id);
```
LEDfan, line 417: Here we can't find the difference between $displayname changed or not changed by the contact owner. This is caused because we don't know the old displayname of the ownCloud user.

tanghus:
@LEDfan Does this need any changes to be merged?

LEDfan:
If you agree with https://github.com/owncloud/contacts/issues/474#issuecomment-42205892. Basically that the FN of the user is always overwritten when a user/admin change the displayname in owncloud.

tanghus:
So the changes made by the users to their contacts that takes precedence like mention in https://github.com/owncloud/contacts/issues/474#issuecomment-41431362 ?
Then I agree - if I haven't misunderstood the question ;)

There are a few style guide issues that you might as well fix while you're there. Mainly always one space after `if`, `foreach`, `while` etc and always one space between `)` and `{`.

LEDfan:
> So the changes made by the users to their contacts that takes precedence like mention in #474 (comment) ?

No, I can't find a way to detect if the user changed\* the FN of the contact, so I can't give it precedence.
- actually there should be a way to compare: 1) the old displayname of the user, 2) the new displayname of the user (thus not set in the contacts app) and 3) the changed FN of the contact app (thus set in the contacts app). AFAIK there isn't a way to find the 1. 

I'll update the styles.

LEDfan:
Implemented in https://github.com/owncloud/core/pull/8951

PR #485, opened 2014-05-08
New import ui dialog

babelouest:
This work is to move the import ui from the bottom-left to a centered dialog window.
testvcard1
Few problems remain, like the upload button which can't be easily moved to a specific location probably due to css rules.

Next step in the todo list will be to let choose the input format with the calculated probabilities if the user selects `Automatic format` in the format list.
1362f9e2-c65c-11e3-8a02-bdc073c1998a


babelouest:
@tanghus , I fixed the problems I mentionned on the above comment, although I have new ones now, as usual :p

tanghus:
@babelouest As is software development ;)

We may wanna get @jancborchardt and @jbtbnl in on this as afairc they had some remarks on the design.

jancborchardt:
If I remember correctly @jbtbnl already gave good feedback on it.

Basically what I think is that import should be simple, and this dialog looks a bit too complicated. It’s just a file I put in there, and the software should solve it intelligently. I should only need to change something / be faced with a dialog if something goes _wrong_.

(Also smaller feedback: Because there’s a closing x in the top right, please do not use additional »Close« or »Cancel« buttons. Label the »OK« button more useful, for example »Import Contacts« – and put it in the bottom right. Also, the icon-only button in the bottom right of the first dialog is not understandable [and the icon is way too big].)

jbtbnl:
Refer to https://github.com/owncloud/contacts/issues/456#issuecomment-40798228

Thanks for creating a pull request :)

babelouest:
@jancborchardt , I understand, that's why I reimplemented the ui as is in the dialog window, because you agreed with this behaviour in the first time (https://github.com/owncloud/contacts/issues/74#issuecomment-33852166 )
According to your previous feedbacks, the remaining problem is the place and look of the upload button, which I agree is not a good one like this. I just can't move it to a different place without breaking its functionnality. So a little help would be appriciated since I'm not very good at web UI.

The css is like this:

``` css
#file_upload_form { display:none; }
#contacts-import-upload-start, .import-upload-button {
    width: 30px !important; height: 30px !important;
display:block;
position: absolute;
right: 0; bottom: 0;
margin: 5px !important;
cursor: pointer;
}
#contacts-import-upload-start {
cursor: pointer;
z-index: 1001;
}
```

If I change the position to float or the coordinates, the button becomes not clickable at all, and I can't figure out why...

According to the `close` button, that's a matter of feeling I would say, I think that the dialog should remain open after the import, to give the user the feedback on the import, let's say he/she has several thousands of contacts to import, that takes time, so he/she goes to something else during that time, and goes back after a random time, if there is import errors or other relevant messages, he/she could not notice because the dialog would have been closed.
And a `close` button that does the same thing than the `x` top-right doesn't shock me, but if you insist to remove it, no problem.

jbtbnl:
@babelouest @jancborchardt what do you think of the following workflow:
- user clicks "Import contacts" => a file chooser window opens for the user to select the file (no dialog window or anything in between)
- show loading animation, the file is uploaded to and parsed by the server 
- The following dialog handles the rest:

contacts import

@babelouest if you'd like I can help you with implementing the UI.

babelouest:
@jbtbnl , thanks a lot, that would be very nice from you to help me implement it. The workflow you describe is very good for me, I like it.
Technically, it means the import backend must have 2 steps at least:
1- upload the file, copy it to the `contacts/upload` folder, parse it to calculate the file format probabilities, then return to the ui a table containing the different formats and their probability
2- import the file to the addressbook using the given format

Although, if we implement it like that, I have a security issue I would like to discuss with you.

Right now, the uploaded file is deleted right after the import. After that, the file will be deleted after the second user action.
If a bloody pirate wants to flood the owncloud server with files, he can send multiple first steps (upload file) and no second steps (import uploaded file).
I know there can be a user quota but still. We could prevent that by forcing the removal of any file in the `data/<user>/contacts/upload` folder before the user uploads a new file.
Of course, it would mean that if multiple users use the same account to upload contacts at the same time, they may have problems but I think it would worth it.

What do you think @tanghus , @jancborchardt , @jbtbnl ?

js/addressbooks.js
```diff
@@ -451,47 +525,22 @@ OC.Contacts = OC.Contacts || {};
...
-	/**
 	 * Create an AddressBook object, save it in internal list and append it's rendered result to the list
 	 *
 	 * @param object addressBook
-	 * @param bool rebuild If true rebuild the address book select for import.
+	 * @param bool rendered If true add the addressbook to the addressbook list
 	 * @return AddressBook
 	 */
-	AddressBookList.prototype.insertAddressBook = function(addressBook) {
+	AddressBookList.prototype.insertAddressBook = function(addressBook, rendered = true) {
```
jbtbnl: You cannot assign a default value in JavaScript this way, and it throws an error. See http://stackoverflow.com/questions/894860/set-a-default-parameter-value-for-a-javascript-function

babelouest: Thanks for the tip, I didn't know.
Although, this function was modified because of underlying bugs I couldn't fix at that time.
In the master branch, there is an up-to-date addressbookList in the javascript objects, but in my case, the list is empty, that's why I use the following process:
Whenever the user clicks on the `import` link, I get all the addressbooks from the backend and adds them into the addressbook list, to fill properly my dropdown list, and I also add the addressbook into the javascript list. But to prevent it from adding addressbooks multiple times in the addressbook list in the top-left corner, I added this hideous hack.
So, after all, the default parameter shouldn't be there because this function shouldn't have been modified in the first place. If one can fix the initial bug, this could solve multiple problems.

jbtbnl: @tanghus can you help @babelouest with this?

jbtbnl:
@babelouest I must admit that I only checked out the branch a couple of hours ago and until then assumed that there was a two-way process implemented like it shows in the screenshots of your first comment https://github.com/owncloud/contacts/pull/485#issue-33048776

Regarding your concern, I see two ways to fix it:
- improve the server security
- do the file analysis at the client side

tanghus:
@babelouest 
Wrt. file upload flood I suggest we simply use the same filename every time. The risk of other users importing from the same account at the same time is close to non-existing imo.
Regarding the addressBookList being doubled I'll try to run through the code later today to see why this could happen.

babelouest:
@tanghus , that's another option, I agree that the risk of multiple users on the same account importing file at the same time is low.
In that case, the filename given in request parameters would be useless except for the upload() function.

Concerning the empty addressBookList when loading the import dialog, I don't figure out why, but since the `ocdialog.open()` function is now called only once, I don't need the line 384:

``` php
self.insertAddressBook(addressBook, false);
```

Although, I can't use the function `self.selectByPermission(OC.PERMISSION_UPDATE);` either, it returns an empty list

tanghus:
> In that case, the filename given in request parameters would be useless except for the upload() function.

Yes, it would. Much of what I've written earlier is ;) We only need the temp file for getting its content.

> Although, I can't use the function self.selectByPermission(OC.PERMISSION_UPDATE); either, it returns an empty list

That should happen only if there are no address books with at least PERMISSION_UPDATE..?

Sorry, things happened so I didn't get a chance to test it this evening. Priority for the weekend.

tanghus:
btw, can I get you to rebase again? I suppose it's https://github.com/owncloud/contacts/commit/f699812f69c214f22bd05ae5ad150daf1da31192 that did it.

jancborchardt:
@babelouest @jbtbnl what’s the status on this by the way? Nicolas can you rebase, so we can test?

babelouest:
@jancborchardt unfortunately, I didn't have time to work on this lately, yet it would be a very good enhancement since it could replace the click on the file app that doesn't work anymore since the new import process...

jbtbnl:
I'll take a look at it next week, self assigning as a reminder

jbtbnl:
@jancborchardt @babelouest still need to work on my merging skills, but I think it is at least a good attempt ;)

jbtbnl:
Todo:
- [x] correct the position of the import dialog to the center of the screen

jbtbnl:
While it does not match the proposed design in https://github.com/owncloud/contacts/pull/485#issuecomment-42554071, I believe this branch is now ready to be merged. Further improvements can be handeled separately.

cc @babelouest @jancborchardt

babelouest:
@jbtbnl , I've just tested your PR and it is awesome. I would add just one thing though, a close button so the user wouldn't get confused and click the upload file button a second time.

Thanks a lot for this !

jbtbnl:
@babelouest good point, I've pushed an additional commit where I hide the import button and select inputs when the import is started. Do you think it is OK like this?

cc @owncloud/designers

stefan-niedermann:
How will this work on mobile devices? I'm confused since kind of these settings and tools should be in every app on the same place - bottom left in our case.

jbtbnl:
@stefan-niedermann the bottom left settings drawer is meant for settings, not for dialogs.

cc @jancborchardt we need a way to display dialogs inline on mobile, because the pop-up kind of dialogs are a pain on small screens, don't you think?

jancborchardt:
On small screens, the popup dialogs should use nearly all of the screen (not all so you still see it’s only a dialog) so that should suffice space-wise.

jancborchardt:
@babelouest @jbtbnl @tanghus @ockham please join the #owncloud-contacts IRC channel I just opened. We should use that to build a stable team around the Contacts app and communicate better.

babelouest:
@jbtbnl , how difficult would it be to adapt this new UI so when on the file app the user selects a vcf or csv or ldif file, it opens your popup with the file selected ?

jbtbnl:
@babelouest UI-wise it is easy, since it is a pop-over. I'm not sure how the integration with files is done though...

babelouest:
This is how I disabled it lately, the code should be adapted somewhere around here: https://github.com/owncloud/contacts/commit/3bb99f30dc7a869a631448225a27fc36bafe3f46

jbtbnl:
@babelouest thanks for pointing me, it is on my todo list now!

But anyway let's rebase this pull request and merge it to at least get this feature in.

babelouest:
:+1:

PR #499, opened 2014-05-20
fix bug #498

babelouest:
A quick change to fix the issue #498.

### DIFF
--- a/lib/addressbook.php
+++ b/lib/addressbook.php
@@ -189,6 +189,8 @@ class Addressbook extends AbstractPIMCollection {
 			$curContact->retrieve();
 			$this->objects[$id] = $curContact;
 			return $curContact;
+		} else {
+			return $this->objects[$id];
 		}
 	}
 
--- a/lib/addressbook.php
+++ b/lib/addressbook.php
@@ -171,11 +171,10 @@ class Addressbook extends AbstractPIMCollection {
 			);
 		}
 
-		if (!isset($this->objects[$id])) {
+		if (!isset($this->objects[(string)$id])) {
 			$contact = $this->backend->getContact($this->getId(), $id);
 			if ($contact) {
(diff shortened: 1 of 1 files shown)


requist:
Tested and works OK here. :+1:

babelouest:
Worked here too with the local and ldap backend, I'm merging

GenXRoad:
Hello, i tested the master, the last dayly build, i even get an errorr on line 64 dispatcher.php so the contacts app will result in 500 internal Server error, which branch will corect  work?

babelouest:
@GenXRoad , are you using the last commit ? I don't get any error with the last one: https://github.com/owncloud/contacts/commit/f6804bac154715cf680804c7d4cb681d510bc651

GenXRoad:
Jes, i' running oc 7 pre alpha since 1 werk dayly upload the new dayly build but since the ldap merg dispatcher.php won't work. I don't know if some files missing? Because clean install oc 7 , the error message i can post later, i'm currenty only with my mobile phone online

LEDfan:
There is a bug in the dispatcher class for PHP 5.3. `$this` should be changed in `$container` Therefore https://github.com/owncloud/contacts/blob/master/lib/dispatcher.php#L87

Refs #481

GenXRoad:
Thank you LEDfan, i will test it today

GenXRoad:
Now it will work, but now an template is missing contacts:formfactor

LEDfan:
@GenXRoad I will fix the PHP 5.3 errors.

LEDfan:
@GenXRoad can you test 0e863592e616a10893c13e26d1844cf1ce21c80b ? It should be solved now.

GenXRoad:
Thank you, tested it localy, will work, on my root i made distro upgrade to newest version with php 5.4

PR #500, opened 2014-05-25
fix plural typo

vgezer:
Has been noticed by [vkehayas](https://www.transifex.com/accounts/profile/vkehayas/) on transifex.


jbtbnl:
:+1:

PR #507, opened 2014-06-08
Change "vcardcountry" to "c" for HOME ADR

makuser:
Instead of "vcardcountry", "c" is used for "RFC2256: ISO-3166 country 2-letter code" in objectClass "inetOrgPerson".

### DIFF
--- a/formats/backend_ldap_inet_org_person_connector.xml
+++ b/formats/backend_ldap_inet_org_person_connector.xml
@@ -42,7 +42,7 @@
 			<ldif_entry vcard_position="3" name="l"/>
 			<ldif_entry vcard_position="4" name="st"/>
 			<ldif_entry vcard_position="5" name="postalcode"/>
-			<ldif_entry vcard_position="6" name="vcardcountry" unique="true"/>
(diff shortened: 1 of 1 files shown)


jbtbnl:
CC @babelouest

babelouest:
I had to go through my openldap schema to figure out why I put this, it's because the 'c' attribute is not present in the inetOrgPerson.schema file, but it exists in the core.schema...
I'll run some tests to validate this, thanks.

jbtbnl:
@babelouest any news on this?

babelouest:
I've tested it in the last days and the change itself seems fine.
I saw potential other problems with the ldap backend, but I couldn't investigate more since I'm moving to a new place right now...

jancborchardt:
@babelouest hope the move went fine! :) Any update on if it could cause problems with LDAP?

@blizzz please help here regarding testing with LDAP.

babelouest:
I think this can be merged anyway, the change is too small to create other bugs I think, and the LDAP backend is still in beta mode so I'm OK with it.

jancborchardt:
Nice, thank you @babelouest! :)

PR #524, opened 2014-06-30
birthday date picker date range

xhess:
change birthday date range from the future to the past.
see: https://github.com/owncloud/contacts/issues/520


jbtbnl:
@xhess thanks for your contribution :+1:

PR #533, opened 2014-07-09
Fix #532. Adds an intermediate variable to be compatible with php 5.3

ghost:


babelouest:
Thanks for the fix, I quickly tested it with my side and it looks fine (Debian Wheezy, php 5.4), you can merge the PR if you want.

babelouest:
:+1:

PR #541, opened 2014-07-18
Fix minor translation mistake

dupondje:


jancborchardt:
@dupondje translations are handled on Transifex: https://www.transifex.com/projects/p/owncloud/ you need to change them directly there. :)

PR #547, opened 2014-07-23
fix php 5.3

BernhardPosselt:
Fix #546 

Not tested, just quick edit


th3fallen:
@Raydiation That isnt going to completely fix his issue, there are two other issues found on lines 195 and 200 the two closures of $writeProgress and $cleanup are unsupported in php5.3 they cannot use $this inside of a closure in that version of php i'm wrapping up a pr to extract them out into methods to resolve the issue.

DeepDiver1975:
I will fix all php 5.3.3 issued in https://github.com/owncloud/contacts/pull/568

PR #561, opened 2014-07-27
Change label when there is no contact

babelouest:
Just a quick fix that was forgotten after the new import process.


jancborchardt:
Looks good! :)

jancborchardt:
Also backported: eef0d2d

PR #563, opened 2014-07-28
fix caldav when the calendar app is only enabled for specific groups

icewind1991:
Since checking if an app is enabled requires a user to be logged in we need to postpone those checks until after the *DAV auth is handled.

Requires owncloud/core#9962

### DIFF
--- a/appinfo/remote.php
+++ b/appinfo/remote.php
@@ -22,7 +22,9 @@
  *
  */
 
-OCP\App::checkAppEnabled('contacts');
+if (!\OC::$server->getAppManager()->isInstalled('contacts')) {
+	throw new Exception('App not installed: contacts');
+}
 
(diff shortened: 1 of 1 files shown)


jancborchardt:
@schiesbn @DeepDiver1975 can you test this, together with https://github.com/owncloud/core/pull/9962

jancborchardt:
@ben-denham @ockham @babelouest can you please review this one?

ockham:
Sorry, I know too little about this to be qualified to review.

babelouest:
@icewind1991 , I understand the first change, but what does the second has to do with this bug ? Is it another way of detecting if the app contact is enabled ?

icewind1991:
Yes, it checks if the app is enabled _after_ webdav auth is done

PVince81:
Works in combination with #9962 :+1:

LukasReschke:
Works :+1:

PR #583, opened 2014-08-07
Backport TagMgr access fix

danimo:
We just noticed that this was missing in stable7 and caused an import error.


danimo:
Wrong target branch

PR #584, opened 2014-08-07
fix TagMgr access remove unwanted error_log

danimo:
We just noticed that this was missing in stable7 and caused an import error.

### DIFF
--- a/lib/controller/importcontroller.php
+++ b/lib/controller/importcontroller.php
@@ -25,9 +25,10 @@ use OCA\Contacts\App,
  */
 class ImportController extends Controller {
 
-	public function __construct($appName, IRequest $request, App $app, ICache $cache) {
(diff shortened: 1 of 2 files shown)


mutax:
It applied to 7.0.1 (stable) and the internal server error on importing is gone. Seems to work. Thanks!

jancborchardt:
Can you review this @babelouest @ockham?

ockham:
I can confirm that on the stable7 branch, importing my test *.csv file fails halfway, while with the backport-fix-TagMgr-access branch it works. So merging is fine with me!

babelouest:
I tested the fix and all my import still pass, merging is fine with me too

ockham:
Okay, merging as there are two positive reviews (including mine) and _because I can_ :-) (I've asked on #owncloud-dev yesterday, but no response. Still hope it's fine.)

Issue #589, opened 2014-08-08
Bad import of oulook csv contact


bernic:
the contact labels get all fusionned together
for example if a contact is labelled Friend; colleague
I will not get 2 labels containing the contact but instead I will get one label "Friend; colleague" containing the contact.
All the labels are scrambled that way, it becomes useless
Regards

babelouest:
closing it due to no response

PR #599, opened 2014-08-10
Fix cacheThumbnail if no $vcard was passed

tribut:
This fixes a type confusion. `getContact` does not return a VCard-object, but an array containing either the VCard mapped to 'vcard' or the string representation of the VCard in 'carddata'.

Note that nowhere in the current codebase is `cacheThumbnail` called without the `$vcard` argument or an argument which could be `null`, so this is not strictly speaking a bug, but dead code which doesn't work.

### DIFF
--- a/lib/utils/properties.php
+++ b/lib/utils/properties.php
@@ -328,6 +328,11 @@ Class Properties {
 			if (is_null($vCard)) {
 				$app = new App();
 				$vCard = $app->getContact($backendName, $addressBookId, $contactId);
+				if (isset($vCard['carddata'])) {
+					$vCard = \Sabre\VObject\Reader::read($vCard['carddata']);
+				} else {
+					$vCard = $vCard['vcard'];
+				}
 			}
 			$image = new \OCP\Image();
 			if (!isset($vCard->PHOTO) || !$image->loadFromBase64((string)$vCard->PHOTO)) {


jancborchardt:
Hey @tribut, thanks a lot for your contribution! :) Can you in turn review some other pull requests here: https://github.com/owncloud/contacts/pulls

@babelouest @ben-denham @ockham please review this :)

jancborchardt:
@tribut by the way, did you hear about the ownCloud Contributor Conf we’re having later this month? See https://owncloud.org/conf :)

babelouest:
I have quickly tested it and I see no obvious problem, looks good to me :)

ockham:
As with #589, I can't say much about this as I don't have any failing use cases that are fixed by this PR. (Which in this  case is kinda obvious if it's about dead code.) Also, I don't know that part of the code really yet.

DeepDiver1975:
@tribut can I ask you to rebase this pr? THX

DeepDiver1975:
@tribut can you also share some background under which conditions this error does pop up?

tribut:
Will have to look into it once again, has been some time since I opened the PR :)
Might be a few days at least, currently I'm pretty busy.

DeepDiver1975:
> Will have to look into it once again, has been some time since I opened the PR :)
> Might be a few days at least, currently I'm pretty busy.

Thanks a lot!

DeepDiver1975:
I just had a deeper look into the code and I seem to miss the necessity of this fix.

tribut:
Closing this as I'm no longer sure this is a sensible fix.

PR #602, opened 2014-08-13
remove obsolete japanese lang code

vgezer:
Similar to https://github.com/owncloud/core/pull/10312.

We only have `ja.php` files. The other one is outdated.


ockham:
+1
I've notice though that there's e.g. de_DE.php (in addition to de.php). What about that?

ockham:
... or pl_PL.php? Looks like there might be some more examples...

vgezer:
Ohh you are right. de_DE is for formal German, but pl_PL and zn_\* should be also removed I think. I will check them out. Thanks for noticing.

MorrisJobke:
:+1:

PR #605, opened 2014-08-18
Import fixes

babelouest:
Fix some bugs that in the import process


babelouest:
should help for #575 , #559

babelouest:
this is in the wrong place, should fix branch stable7, not master (it should also be backported into master, but after)

PR #612, opened 2014-08-22
Fixes Issue #301

acrylschaf:
Gives even Windows Phone 8 users the chance to celebrate their loved
ones' birthdays. https://github.com/owncloud/contacts/issues/301

### DIFF
--- a/lib/contact.php
+++ b/lib/contact.php
@@ -827,7 +827,7 @@ class Contact extends VObject\VCard implements IPIMObject {
 			);
 			$vevent->add('DURATION', 'P1D');
 			$vevent->{'UID'} = $this->UID;
-			$vevent->{'RRULE'} = 'FREQ=YEARLY';
+			$vevent->{'RRULE'} = 'FREQ=YEARLY;BYDAY=MO,TU,WE,TH,FR,SA,SU;BYMONTH=' . $date->format('n') . ';BYMONTHDAY=' . $date->format('j') . ';INTERVAL=1';
 			$vevent->{'SUMMARY'} = $title . ' (' . $date->format('Y') . ')';
 			$vcal = \Sabre\VObject\Component::create('VCALENDAR');
 			$vcal->VERSION = '2.0';


jbtbnl:
cc @babelouest for his opinion

babelouest:
Does that involve a change in the non-windows vcards ?
If so, I'm not sure it's a good idea to add and maintain a hack just for one client app workaround.

acrylschaf:
Is it even allowed according to the standard to leave "FREQ=YEARLY" as only tag in RRULE (as ownCloud does it)? Looking through http://www.ietf.org/rfc/rfc4791.txt I can only find yearly recurrences along with the BYMONTH and BYDAY tag.
That's also the way entries are formatted on Windows Phone.

In any way, the extra tags would make the calendar scheduling simpler for every client (which leads to less errors in the end).

acrylschaf:
Would be awesome if the fix was accepted - it already was by the calendar team: https://github.com/owncloud/calendar/pull/536
I assume also other client apps (like older versions of iCal?) can't handle owncloud's syntax for yearly-reoccuring events: http://forum.owncloud.org/viewtopic.php?f=14&t=21377

DecaTec:
Is this fixed in OC 7.04? I stumbled upon the issue with my WinPhone and it doesn't seem to work.

jbtbnl:
Since this change should be according to spec and as reported solves an issue with certain clients, I'd say we better merge this :+1: 

cc @georgehrke since he merged https://github.com/owncloud/calendar/pull/536

DeepDiver1975:
@jakobsack can you please review this - I don't yet have an understanding about the vcard stuff - THX

jakobsack:
The new syntax seems to be fine. If it helps we can merge it.

As a side note: the current RRULE is valid according to the RFC and my Windows Phone 8.1 already shows all birthdays.

georgehrke:
`FREQ=YEARLY` is a perfectly valid value for `RRULE`. So is the new value, but if Windows phone 8.1 fixed it, I would actually refrain from making the `RRULE` overly complex.

jakobsack:
I agree with Georg. Additionally, _all_ Windows 8 phones should be able to upgrade to 8.1.

To quote the RFC 5545 (3.3.10):

```
  Information, not contained in the rule, necessary to determine the
  various recurrence instance start time and dates are derived from
  the Start Time ("DTSTART") component attribute.  For example,
  "FREQ=YEARLY;BYMONTH=1" doesn't specify a specific day within the
  month or a time.  This information would be the same as what is
  specified for "DTSTART".
```

jancborchardt:
So what’s the call on this change @jakobsack @georgehrke merge or not?

georgehrke:
I'd vote to close it.

jakobsack:
close it

PR #621, opened 2014-08-29
Move creation of default address book to backend

### LINKED ISSUE #513: navigation with tab key buggy
### Steps to reproduce
1. Open contacts app in your nextcloud
2. Click on any existing contact / Or add a new contact
3. Navigate through the input fileds using your tab key
### Expected behaviour

One would expect to navigate from field directly to the next field - exept beside the fileds are other form elements which allow to take action on something.
### Actual behaviour
1. When the cursor is in the field "city" (within adress) and I press tab it jumps to the label of the next field "state / province" and not directly to the input field.
2. When the cursor is in the field "firstname" (within detailed / full name) and I press tab it jumps to the label of the next field "additional name" and not directly to the input field.
3. It is also possible to navigate to the label "detailed name" above all name related input fields.
### Server configuration

**Operating system**: Debian 3.10

**Web server:** Apache 2.4.10

**Database:** MySQL 15.1 Distrib 10.0.26-MariaDB

**PHP version:** PHP 5.6.24-0

**Nextcloud version:** 10.0

**Updated from an older Nextcloud/ownCloud or fresh install:** Updated since owncloud 7

**Where did you install Nextcloud from:** github/nextcloud/server

jakobsack:
Fixes owncloud/contacts#513

Works very well on my system, but localization is not working when accessing the app via carddav.

### DIFF
--- a/lib/app.php
+++ b/lib/app.php
@@ -105,20 +105,6 @@ class App {
 			foreach (array_keys(self::$backendClasses) as $backendName) {
 				$backend = self::getBackend($backendName, $this->user);
 				$addressBooks = $backend->getAddressBooksForUser();
-				if ($backendName === 'local' && count($addressBooks) === 0) {
-					$id = $backend->createAddressBook(array('displayname' => self::$l10n->t('Contacts')));
-					if ($id !== false) {
-						$addressBook = $backend->getAddressBook($id);
-						$addressBooks = array($addressBook);
-					} else {
-						\OCP\Util::writeLog(
-							'contacts',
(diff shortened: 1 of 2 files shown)


jbtbnl:
@schnied please post your findings with this fix here, so that we know that it is working for you. Thanks!

schnied:
sorry for my late answer.
yes, this fix works fine for me.

Am 13.09.14 um 16:37 schrieb jbtbnl:

LEDfan:
I didn't test the originally issue, but @schnied says it solved. I tested and addressbook are still created.

:+1:

PR #653, opened 2014-09-16
Contacts: Issue #600 partial fix.

synchrone:
It can sometimes be caused by the search provider ajax result set formatting issue, since the arguments are not in the correct order or the display name being null or an empty string depending on the user data, which leads to an unselectable row in search results

### DIFF
--- a/lib/searchprovider.php
+++ b/lib/searchprovider.php
@@ -12,29 +12,25 @@ namespace OCA\Contacts;
 
 class SearchProvider extends \OC_Search_Provider{
 	public function search($query) {
-		$unescape = function($value) {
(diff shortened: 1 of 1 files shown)


butonic:
obsoleted by 
7d8fe723514c8dd329340db2d7a3a66a0c448626
e426d7e8679ccd00c058e61bb2a85da3584665fa
7c54ac4f755d0df5e45a3fdbb14949b9d8767d1e
ea7f1545a7ad101393fb745406efadeb0aa7e83d

they also update the used classes to the public API. Please have a look at how to further change the search result display in https://github.com/owncloud/contacts/blob/master/js/loader.js#L136

PR #665, opened 2014-09-29
Share contact groups

ockham:
This PR depends on owncloud/core#11351 and is supposed to fix #38.


ockham:
D'oh. I'll rebase against master :-/

PR #666, opened 2014-09-30
Update contacts.php

master-pago:
Removed avatar icon (line 185) because it was creating enourmous extra loading time (ca. +1000%)


jbtbnl:
:-1: we need lazy loading for this https://github.com/owncloud/contacts/issues/582#issuecomment-58051769

PR #667, opened 2014-10-01
Share contact groups2

ockham:
Supersedes #665; depends on owncloud/core#11383. Should fix #38.


ockham:
Had to rebase, see #668.

PR #668, opened 2014-10-01
Share contact groups2

ockham:
Supersedes #665; depends on owncloud/core#11383. Should fix #38.

### DIFF
--- a/lib/controller/groupcontroller.php
+++ b/lib/controller/groupcontroller.php
@@ -65,7 +65,7 @@ class GroupController extends Controller {
 
 		$response = new JSONResponse();
 
-		if (is_null($name) || $name === "") {
+		if (empty($name)) {
 			$response->bailOut(App::$l10n->t('No group name given.'));
 		}
 
@@ -88,7 +88,7 @@ class GroupController extends Controller {
 		$name = $this->request->post['name'];
 
 		$response = new JSONResponse();
-		if (is_null($id) || $id === '') {
+		if (empty($id)) {
 			$response->bailOut(App::$l10n->t('No group ID given.'));
 			return $response;
 		}
@@ -148,12 +148,12 @@ class GroupController extends Controller {
 
 		$response = new JSONResponse();
 
-		if (is_null($from) || $from === '') {
+		if (empty($from)) {
 			$response->bailOut(App::$l10n->t('No group name to rename from given.'));
 			return $response;
 		}
 
-		if (is_null($to) || $to === '') {
+		if (empty($to)) {
 			$response->bailOut(App::$l10n->t('No group name to rename to given.'));
 			return $response;
 		}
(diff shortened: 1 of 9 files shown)


jbtbnl:
@ockham great to see that you are working on this!

cc @babelouest @LEDfan

ockham:
cc @LukasReschke @MorrisJobke @Raydiation

lib/controller/groupcontroller.php
```diff
@@ -83,22 +84,24 @@ public function addGroup() {
 	 * @NoAdminRequired
 	 */
 	public function deleteGroup() {
+		$id = $this->request->post['id'];
 		$name = $this->request->post['name'];
 
 		$response = new JSONResponse();
-		if (is_null($name) || $name === '') {
-			$response->bailOut(App::$l10n->t('No group name given.'));
+		if (is_null($id) || $id === '') {
```
LukasReschke: is_empty?

ockham:
CC'ing @wakeup as you're one of the more active people here in contacts.

ockham:
Everybody, please consider reviewing this PR even if you feel you aren't that familiar with it. 
I'm afraid there aren't really that many people around (or active) that really know the Contacts app code that well, so it's up to those that don't ;-)

As for the commits, most of them are either fixes to currently somewhat dead parts of the code that this feature requires, or they enable operations (think CRUD) based on group id instead of name; the most part of the changes to the JavaScript files, and obviously the GroupController, is about this capability.

It's probably a good idea to just fetch the branch and try it out to get an idea on what it does. 
To that end, on a local installation with at least two users, both should have some contacts and contact groups, and one of them needs to share their addressbook with the other. Then, log in as the other user, and play with the shared contacts groups, i.e. try creating, renaming, deleting them.

jbtbnl:
@ockham I gave it a spin, and have the following remarks:
- [ ] the number of contacts is incorrectly displayed as 1
- [ ] editing the address book from the user account it is shared with, results in the LDAP edit dialog

ockham:
@jbtbnl Thank you very much for testing!

About those issues you spotted, I think they're also present in the current master branch and 7.0.x releases, i.e. not related to my patch. Can you confirm?

jbtbnl:
@ockham ah, yes indeed! Those issues are in the master branch. I found a related issue for the first one: https://github.com/owncloud/contacts/issues/654.

Anyway, :+1: for this PR

jbtbnl:
@ockham shall we merge this just as https://github.com/owncloud/contacts/pull/686#issuecomment-62204688?

ockham:
Yeah.

cfggns:
The issue this PR is supposed to fix (#38) still seems to be in a problem on the latest version of ownCloud. I see the PR has been merged, but is there some reason why it may not have been incorporated into the ownCloud server packages? It would be really  useful to be able to share contact groups!

PR #682, opened 2014-10-19
Fix circular structure typeerror in chromium(38)

seijikun:
The variable `params` contains the member `obj` which does not seem to be of interest by the server. In chromium, this obj - member leads to an `Uncaught TypeError: Converting circular structure to JSON` when calling `JSON.stringify` for `params` in the `storage.patchContact(...)` function. There might be some other points within the code with the same problem, but I didn't find any on the fly.


jancborchardt:
@DeepDiver1975 @babelouest @LEDfan what’s the call here?

DeepDiver1975:
:+1:

PR #686, opened 2014-10-28
Display correct number of contacts in shared addressbooks.

ockham:
Fixes #654.

Please review @jbtbnl @nexusnetwork23 !

### DIFF
--- a/js/groups.js
+++ b/js/groups.js
@@ -751,7 +751,7 @@ OC.Contacts = OC.Contacts || {};
 					$elem = $elem.length ? $elem : (tmpl).octemplate({
 						id: shared.id,
 						type: 'shared',
-						num: response.data.shared.length,
+						num: shared.length,
 						name: shared.displayname
 					});
 					$elem.find('.numcontacts').after(sharedindicator);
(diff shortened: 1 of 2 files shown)


jbtbnl:
Nice! Tested, works well :+1:

ockham:
@jbtbnl Thx for the review!

Let's CC some more folks: @babelouest @LEDfan @AnybodyElse @Raydiation @DeepDiver1975

ockham:
CCing @MorrisJobke @LukasReschke 

Simple bug fix here!

ockham:
Given that currently it's very hard to attract more reviewers, and following discussion on `#owncloud-dev` where it was stated that two reviewers were a hard requirement only for core but not necessarily for apps as long as all app maintainers agreed, I'm merging this.

As the Contacts app currently doesn't really have any actual maintainers, with only @jbtbnl (who has :+1:'d this PR) and myself being the most active devs, I take it as concludent that it's okay to do so.

jbtbnl:
@ockham totally agree!

MorrisJobke:
@ockham Yes. Thanks for your work :)

PR #696, opened 2014-11-07
Load md5 and Jcrop scripts and styles from their new locations.

ockham:
They could not be loaded as they have been replaced by their bower
counterparts with
owncloud/core@49fbd766c2ad6e5bc7c8401bfab2c20b1f442012 and
owncloud/core@73569b29bc3984e1e6df102faccb3c02c20573b3

### DIFF
--- a/templates/contacts.php
+++ b/templates/contacts.php
@@ -1,5 +1,43 @@
 <?php
 use OCA\Contacts\ImportManager;
+
+script('', [
+	'jquery.multiselect',
+	'tags',
+	'placeholder',
+	'jquery.avatar',
+	'avatar',
+]);
+
+vendor_script('', [
+	'blueimp-md5/js/md5',
+	'jcrop/js/jquery.Jcrop',
+]);
+
+script('files', 'jquery.fileupload');
+
+script('contacts', [
+	'jquery.combobox',
+	'modernizr.custom',
+	'app',
+	'addressbooks',
+	'contacts',
+	'storage',
+	'groups',
+	'jquery.ocaddnew',
+	'otherbackendconfig',
+]);
+
+style('', 'jquery.multiselect');
+
+vendor_style('jcrop/css/jquery.Jcrop');
+
+style('contacts', [
+	'jquery.combobox',
+	'jquery.ocaddnew',
+	'contacts',
+]);
+
 ?>
 <div id="app">
 	<div id="app-navigation" class="loading">
--- a/lib/controller/pagecontroller.php
+++ b/lib/controller/pagecontroller.php
@@ -42,29 +42,6 @@ class PageController extends Controller {
 
 		$maxUploadFilesize = \OCP\Util::maxUploadFilesize('/');
 
-		\OCP\Util::addScript('', 'jquery.multiselect');
-		\OCP\Util::addScript('', 'tags');
-		\OCP\Util::addScript('placeholder');
-		\OC_Util::addVendorScript('blueimp-md5/js/md5');
-		\OCP\Util::addScript('jquery.avatar');
-		\OCP\Util::addScript('avatar');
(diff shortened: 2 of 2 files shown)


ockham:
Reviews, pls!

@Raydiation you mentioned this today around noon on `#owncloud-dev`
@jbtbnl @LukasReschke @MorrisJobke

ockham:
CC'ing @babelouest @LEDfan and @wakeup

jbtbnl:
@ockham is there a reason to switch to OC_Util? Is either one depreciated?

ockham:
addVendorScript and addVendorStyle are only there, but not in \OCP\Util

ockham:
alternative would be to use something like `\OCP\Util::addScript('vendor/blueimp-md5', 'md5');`, but addVendorScript sounded more to the point (it's what @MorrisJobke used in the commits I linked to)

ockham:
Hm, technically, `OC_Util` is private. So maybe we should just add those two functions to the `OCP\Util`.

jbtbnl:
OK, let's wait for a reply from one of the core developers then to hear what's the preferred method...

ockham:
Okay, filed a PR: https://github.com/owncloud/core/pull/12039

jbtbnl:
While I'm thinking about it, is it preferred to rely on libraries in core? It is clear that core developers are moving JS dependencies around, but it would be good to have a heads-up on devel@owncloud.org about this.

cc @MorrisJobke

BernhardPosselt:
Check how the news app does it, you should load the scripts in the templates

ockham:
@Raydiation Can I use the `style` and `script` commands for js and css files that are in core, too, or do I need to ship everything redundantly with the contacts app? (BT Are there any docs for those commands?)

BernhardPosselt:
@ockham you should ship css and js files separately to prevent breakage like this in the future

ockham:
Hm. I think there should really be some defined set of standard JS libraries along with their CSS files so apps don't have to duplicate them needlessly.

BernhardPosselt:
Its a few kbs and youre in control of what you ship which is very convenient. It is actually the declared goal that your app is self contained, meaning that it has as little deps on core as needed

MorrisJobke:
> @ockham is there a reason to switch to OC_Util? Is either one depreciated?

We're not finished with the transition and plan to move all the CSS and JS loading stuff from OC_Util to a more template related code path. Therefore we don't introduced the new public API because it will most likely removed in some days/weeks

ockham:
Fair enough. I've installed bower and will use it to download those scripts into a `vendor`subdirectory.
Still, I would like to merge this PR until then, as I don't want to fix one issue with a PR that entails a lot of possible other issues (like using new versions etc.)

I've rolled back the second commit, so we don't depend on the availability of `addVendorStyle` and `addVendorScript` in `\OCP\Util` anymore, but can just apply it as-is.

ockham:
The latest commit ea7e3b4 now moves loading of scripts and styles to the `contacts.php` template. Consider this an intermediate stage before shipping our own versions of the vendor scripts.

BernhardPosselt:
Indeed, already looks pretty good, :+1: from my side

jbtbnl:
@ockham yes, let's fix master for now :+1:

templates/contacts.php
```diff
@@ -1,5 +1,43 @@
 <?php
 use OCA\Contacts\ImportManager;
+
+script('', [
```
MorrisJobke, line 4: You could simply omit the empty string AFAIK

ockham, line 4: I tried it - it doesn't work :-(

MorrisJobke, line 4: Okay. Then I mixed this up :(

PR #703, opened 2014-11-15
User bower to load JS dependencies + don't use JS from core or other apps

LEDfan:
I think this is earlier discussed in https://github.com/owncloud/contacts/pull/696. This also fixs https://github.com/owncloud/contacts/issues/702 .

Please review @Raydiation @jbtbnl @ockham


BernhardPosselt:
:+1:

ockham:
Doesn't work for me (yet): when clicking on a contact, the edit form should appear, but it doesn't. :-1: 
Console error log (reported for line 1098 in contacts.js):

```
TypeError: self.$groupSelect.multiselect is not a function
```

LEDfan:
@Ockham can you test with  d13aa6b

ockham:
@LEDfan :frowning: I'm still getting the same error.

jbtbnl:
@LEDfan you've got the wrong multiselect library :wink: 

We need https://github.com/ehynds/jquery-ui-multiselect-widget

jbtbnl:
https://github.com/owncloud/contacts/commit/0b36785bbee702dd6fae6dbc9e426f40e4205acd should fix it @ockham

ockham:
Yeah, it does. Great!
Hm, would you mind squashing the last two commits into the ones where they belong before we merge?

jbtbnl:
I just read the mail of @MorrisJobke:

> We also introduce two new template shortcuts (pending documentation PR can be found at https://github.com/owncloud/documentation/pull/636):
> 
> vendor_script(APP, FILE)
> vendor_style(APP, FILE)
> 
> They both result in path/to/apps/APP/vendor/FILE.js (or .css). If APP is omitted they will be looked up in owncloud/core/vendor/FILE.js. The core PR for these functions is #11916 and is already merged into master. 

So I guess we should change the path from `/js/bower_components` to `/vendor`

LEDfan:
@jbtbnl you are right about the path.

ockham:
Yup.

PR #706, opened 2014-11-17
Fix the createOrUpdate function

LEDfan:
As said in #704, this function doens't work and I even think it never has worked... (or for a short time)
This function is accessed by the ContactsManager in core. 

cc @DeepDiver1975

### DIFF
--- a/lib/addressbookprovider.php
+++ b/lib/addressbookprovider.php
@@ -225,7 +225,7 @@ SQL;
 		$contact->save();
 
 
-		return $this->getAddressbook()->getChild($contact->getId());
+		return $contact;
 	}
 
 	/**
--- a/lib/addressbookprovider.php
+++ b/lib/addressbookprovider.php
@@ -178,51 +178,35 @@ SQL;
 	* @return mixed
 	*/
 	public function createOrUpdate($properties) {
-		$id = null;
 
-		/**
-		 * @var \OCA\Contacts\VObject\VCard
-		 */
-		$vcard = null;
-		if(array_key_exists('id', $properties)) {
-			// TODO: test if $id belongs to this addressbook
-			$id = $properties['id'];
-			// TODO: Test $vcard
-			$vcard = $this->addressBook->getChild($properties['id']);
-			foreach(array_keys($properties) as $name) {
-				if(isset($vcard->{$name})) {
-					unset($vcard->{$name});
-				}
-			}
-		} else {
-			$vcard = \Sabre\VObject\Component::create('VCARD');
-			$uid = substr(md5(rand().time()), 0, 10);
-			$vcard->add('UID', $uid);
-			try {
-				$id = $this->addressBook->addChild($vcard);
-			} catch(\Exception $e) {
-				\OCP\Util::writeLog('contacts', __METHOD__ . ' ' . $e->getMessage(), \OCP\Util::ERROR);
-				return false;
-			}
-		}
+        $addressBook = $this->getAddressbook();
 
-		foreach($properties as $name => $value) {
+        try {
+            $id = $addressBook->addChild();
+        } catch(\Exception $e) {
+        }
+
+        if ($id === false) {
+        }
+
+        $contact = $addressBook->getChild($id);
+        foreach($properties as $name => $value) {
 			switch($name) {
 				case 'ADR':
 				case 'N':
 					if(is_array($value)) {
 						$property = \Sabre\VObject\Property::create($name);
-						$property->setParts($value);
-						$vcard->add($property);
+						$property->setValue($value);
+						$contact->add($property);
 					} else {
-						$vcard->{$name} = $value;
+						$contact->{$name} = $value;
 					}
 					break;
 				case 'BDAY':
 					// TODO: try/catch
 					$date = New \DateTime($value);
-					$vcard->BDAY = $date->format('Y-m-d');
-					$vcard->BDAY->VALUE = 'DATE';
+					$contact->BDAY = $date->format('Y-m-d');
+					$contact->BDAY->VALUE = 'DATE';
 					break;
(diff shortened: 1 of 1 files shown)


lib/addressbookprovider.php
```diff
@@ -139,51 +139,35 @@ public function search($pattern, $searchProperties, $options) {
...
-		}
+        $addressBook = $this->getAddressbook();
+
+        try {
+            $id = $addressBook->addChild();
+        } catch(\Exception $e) {
+        }
 
-		foreach($properties as $name => $value) {
+        if ($id === false) {
```
jbtbnl: @LEDfan what for?

lib/addressbookprovider.php
```diff
@@ -139,51 +139,35 @@ public function search($pattern, $searchProperties, $options) {
...
-			$uid = substr(md5(rand().time()), 0, 10);
-			$vcard->add('UID', $uid);
-			try {
-				$id = $this->addressBook->addChild($vcard);
-			} catch(\Exception $e) {
-				\OCP\Util::writeLog('contacts', __METHOD__ . ' ' . $e->getMessage(), \OCP\Util::ERROR);
-				return false;
-			}
-		}
+        $addressBook = $this->getAddressbook();
```
jbtbnl: @LEDfan please fix the indentation, and use only tabs

lib/addressbookprovider.php
```diff
@@ -139,51 +139,35 @@ public function search($pattern, $searchProperties, $options) {
...
+
+        $contact = $addressBook->getChild($id);
+        foreach($properties as $name => $value) {
 			switch($name) {
 				case 'ADR':
 				case 'N':
 					if(is_array($value)) {
 						$property = \Sabre\VObject\Property::create($name);
-						$property->setParts($value);
-						$vcard->add($property);
```
jbtbnl, line 216: Since the method returns a Card object (https://github.com/owncloud/contacts/blob/fix_704/lib/carddav/addressbook.php#L210) it wouldn't be so bad to have it named $vcard

lib/addressbookprovider.php
```diff
@@ -191,27 +175,21 @@ public function createOrUpdate($properties) {
...
-		} catch(\Exception $e) {
-			\OCP\Util::writeLog('contacts', __METHOD__ . ' ' . $e->getMessage(), \OCP\Util::ERROR);
-			return false;
-		}
-		
-		$asarray = VCard::structureContact($vcard);
-		$asarray['id'] = $id;
-		return $asarray;
+
+		return $this->getAddressbook()->getChild($contact->getId());
```
jbtbnl: I suppose it could directly return $contact? It would save another database query

lib/addressbookprovider.php
LEDfan, line 216: The `getAddressbook` function returns `OCA\Contacts\Addressbook` (https://github.com/owncloud/contacts/blob/fix_704/lib/addressbook.php#L167) The getChild function returns a `OCA\Chat\Contact` object.

lib/addressbookprovider.php
LEDfan: Fixed.

lib/addressbookprovider.php
LEDfan: Fixed.

lib/addressbookprovider.php
jbtbnl, line 216: Ah, yes you're right :)

jbtbnl:
@LEDfan the "OrUpdate" part still needs to be taken care of, right? Atm  there is a new child inserted to the address book at every call.

lib/addressbookprovider.php
```diff
@@ -139,51 +139,32 @@ public function search($pattern, $searchProperties, $options) {
...
+		$addressBook = $this->getAddressbook();
+
+		try {
+			$id = $addressBook->addChild();
+		} catch(\Exception $e) {
 		}
 
+		$contact = $addressBook->getChild($id);
 		foreach($properties as $name => $value) {
 			switch($name) {
```
jbtbnl: Instead of this switch, a call to [$contact->setPropertyByName($name, $value)](https://github.com/owncloud/contacts/blob/master/lib/contact.php#L562) will probably work and avoid code duplication. (it's not your code, but we'd better get it right this time :) )

LEDfan: Fixed :)

LEDfan:
@jbtbnl can you review this please? :)

lib/addressbookprovider.php
```diff
@@ -178,79 +178,32 @@ public function search($pattern, $searchProperties, $options) {
...
 			$id = $properties['id'];
-			// TODO: Test $vcard
-			$vcard = $this->addressBook->getChild($properties['id']);
+			$contact = $this->addressBook->getChild($properties['id']);
 			foreach(array_keys($properties) as $name) {
-				if(isset($vcard->{$name})) {
-					unset($vcard->{$name});
-				}
+				if(isset($contact->{$name})) {
+					unset($contact->{$name});
```
jbtbnl, line 189: If this necessary? https://github.com/owncloud/contacts/pull/706/files#diff-c28224c652336b536e90c0435747853aR203 should be able to overwrite existing properties, doesn't it?

jbtbnl:
@LEDfan way better, and less code! As I already commented on the diff, please check if removing the foreach (plus body) on https://github.com/owncloud/contacts/blob/fix_704/lib/addressbookprovider.php#L187 is possible.

lib/addressbookprovider.php
LEDfan, line 189: I tested without, sometimes it works but sometimes there is strange behaviour. (Like not overwriting etc) I would keep it.

PR #710, opened 2014-11-21
Fix 404 contact not found with shared addressbook

LEDfan:
There are again 404 occuring, this time with a shared addressbook. The contacts app tries to search for a contact in a `local` addressbook, but it fails because the addressbook isn't owned by this user. Instead it should access the same addressbook but not by using the `local` backend but the `shared` backend.

Refs https://github.com/owncloud/chat/issues/159#issuecomment-63945487

Note that this solution is a hack! In the DB there isn't a field which backend is used for this contact, thus we should guess this. Luckily the contact id is unique -> there can't be a match in the local addressbook when the contact is stored in the shared addressbook.

### DIFF
--- a/lib/addressbookprovider.php
+++ b/lib/addressbookprovider.php
@@ -287,18 +287,4 @@ SQL;
 		return $result;
 	}
 
-	private function getAddresbookKeyForContact($id){
-		$table = self::CONTACT_TABLE;
-		$query = <<<SQL
-			SELECT
-				`addressbookid`
-			FROM
-				`$table`
-			WHERE `id` = ?
-SQL;
-		$stmt = \OCP\DB::prepare($query);
-		$result = $stmt->execute(array($id));
-		$row = $result->fetchOne();
-		return $row['addressbookid'];
-	}
 }
--- a/lib/addressbookprovider.php
+++ b/lib/addressbookprovider.php
@@ -110,39 +110,31 @@ class AddressbookProvider implements \OCP\IAddressBook {
 			return false;
 		}
 		while( $row = $result->fetchRow()) {
-			$contacts[] = array(
-				"id" => $row['contactid'],
-				"addressbook_key" => $this->getAddresbookKeyForContact($row['contactid'])
-			);
-		}
-
-		if(count($contacts) > 0) {
-			foreach($contacts as $id){
-				try {
-					// gues that it is a local addressbook
-					$contact = $this->app->getContact('local', $id['addressbook_key'], $id['id']);
-				} catch (\Exception $e) {
-					if ($e->getCode() === 404){
-						// not a local thus it is a shared
-						$contact = $this->app->getContact('shared', $id['addressbook_key'], $id['id']);
-					}
+			$id = $row['contactid'];
+			$addressbookKey = $this->getAddresbookKeyForContact($row['contactid']);
+			try {
+				// gues that it is a local addressbook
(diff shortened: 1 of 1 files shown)


LEDfan:
cc @babelouest @jbtbnl @DeepDiver1975

lib/addressbookprovider.php
```diff
@@ -266,4 +271,19 @@ private function convertToSearchResult($j) {
 
 		return $result;
 	}
+
+	private function getAddresbookKeyForContact($id){
+		$table = self::CONTACT_TABLE;
+		$query = <<<SQL
```
LEDfan: I'll add this to the main query.

LEDfan: Done

lib/addressbookprovider.php
```diff
@@ -187,6 +206,7 @@ public function createOrUpdate($properties) {
 					break;
 				case 'EMAIL':
 				case 'TEL':
+				case 'TEL':
```
jbtbnl: I think this is a typo :)

LEDfan: Indeed, thanks!

Gomez:
:+1: Works now! Get it in stable7 isn't a problem, right?

lib/addressbookprovider.php
```diff
@@ -89,11 +90,26 @@ public function getPermissions() {
...
+				`$propTable`.`contactid`,
+				`$contTable`.`addressbookid`
+			FROM
+				`$propTable`
+			INNER JOIN
+				`$contTable`
+			ON `$contTable`.`id` = `$propTable`.`contactid`
+			WHERE
+			  `$propTable`.`userid` = ?
+			AND (
```
jbtbnl: Whats up with this AND?

LEDfan:
@Gomez :fireworks: :fireworks: Nice! I'll backport it for 7.0.4.

jbtbnl:
@LEDfan let's wait for a go from @babelouest before backporting :)

lib/addressbookprovider.php
LEDfan: It's used by lines 112-116, to add the search parameters.

jbtbnl: Ah I see, thanks

LEDfan:
Please don't merge, since there is still something wrong with this solution.

LEDfan:
I added another commit so the shared contacts are returned too.

@Gomez can you test if it still works? Tnx

Gomez:
@LEDfan Works fine. And i think this fixed the contacts count on shared addressbooks, too.

babelouest:
I haven't test it yet, but the code looks clean.
I will test it with the ldap backend too, but I'm not even sure if the search engine works with this backend :p

Gomez:
Can i do something further to get it merged?

Gomez:
Still possible to get this in 7.0.4? Would be good!

LEDfan:
@babelouest @jbtbnl can this be merged and backported to 7.0.4?

babelouest:
I have problems using owncloud today: #724 , I will try to test it as soon as I can

babelouest:
That works for me, a shared contact is searchable, no 404, although search doesn't work with ldap backend, this can be fix later.

Gomez:
Oh, can we backport into stable7

Gomez:
@LEDfan Can we backport to stable7? I would use this branch on my installations, to bad it hasnt got into 7.0.4 :(

PR #714, opened 2014-11-24
Remove OC_Migrate provider

LukasReschke:
See https://github.com/owncloud/core/pull/12380


jbtbnl:
:+1: thanks for the PR!

PR #725, opened 2014-12-02
make search results app specific

butonic:
requires owncloud/core#12555 to be merged, will make this app require oc8


LEDfan:
I think the master branch is the ownCloud 8.0 version, so shouldn't be a problem. cc @babelouest @jbtbnl

PVince81:
:+1:

PR #728, opened 2014-12-05
Fix vendor_script not found

LEDfan:
Fixs #727 
Note that this issue only happens when you are using `master` contacts and `stable7` ownCloud.

### DIFF
--- a/lib/compat.php
+++ b/lib/compat.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * Copyright (c) 2014, Tobia De Koninck hey--at--ledfan.be
+ * This file is licensed under the AGPL version 3 or later.
+ * See the COPYING file.
+ */
+
+if(!function_exists('vendor_script')){
+	function vendor_script($app, $files){
+		if(is_array($files)) {
+			foreach($files as $file) {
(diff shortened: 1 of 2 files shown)


jbtbnl:
cc @Raydiation @MorrisJobke, I don't think it is desirable to have such patches in apps. Either core should provide a fallback, or we should accept that the master branch is incompatible with stable7, which is likely the case anyway.

BernhardPosselt:
Doing the same here: https://github.com/owncloud/news/blob/master/templates/index.php#L12

:+1:  from my side :)

BernhardPosselt:
You could ofc create a separate app which includes backports but then all hell will break loose with people going ermagerd noes app derperndencies

jbtbnl:
@LEDfan this PR conflicts with your point of view at https://github.com/owncloud/contacts/pull/725#issuecomment-65752970

LEDfan:
@jbtbnl you are right. I created this PR because someone in the mailing list was using OC7.0.3 and Contacts master

jbtbnl:
@LEDfan yeah, anytime soon there should be support in Core for specifying dependencies for apps. Until then we'll stick to single version compatibility and backports for bugs.

PR #729, opened 2014-12-05
Use correct combobox

LEDfan:
I added the wrong combobox to bower.....
This is the result:
screen shot 2014-12-05 at 08 29 26

And with this PR, it's again like it was before:
screen shot 2014-12-05 at 08 30 01


jbtbnl:
@LEDfan nice find! :+1:

PR #734, opened 2014-12-08
fix auto escaping functionality of t()

MorrisJobke:
Because of https://github.com/owncloud/core/pull/12687

Fully backwards compatible format. Once all supported platforms (that this version of the contacts app should run on) are using the autoescaping, this can be done via the build in auto escaping functionality of t().

cc @babelouest @jbtbnl

@LukasReschke @PVince81 @karlitschek Example of fully backwards compatible version for this change

### DIFF
--- a/js/groups.js
+++ b/js/groups.js
@@ -530,7 +530,7 @@ OC.Contacts = OC.Contacts || {};
 				if(self.hasGroup(newname, owner)) {
 					$(document).trigger('status.contacts.error', {
 						error: true,
-						message: t('contacts', 'A group named "{group}" already exists', {group: escapeHTML(newname)})
+						message: t('contacts', 'A group named "{group}" already exists', {group: newname})
 					});
 					return;
 				}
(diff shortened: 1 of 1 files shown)


jbtbnl:
@MorrisJobke can we do
`t('contacts', 'A group named "{group}" already exists', {group: newname})`
instead of
`t('contacts', 'A group named "{group}" already exists', {group: escapeHTML(newname)}, null, true)`
?

MorrisJobke:
@jbtbnl Sure, but then it isn't combatible to stable7 and stable6. I don't know how this is handled here, but I thought there is just one version of contacts that works on stable6, 7 and master.

jbtbnl:
@MorrisJobke contacts has stable branches, just as core :)

The master branch of Contacts is only compatible with core master. Otherwise it would need a lot of legacy code and oC version dependent fixes.

MorrisJobke:
@jbtbnl adopted for the updated core PR

Issue #488, opened 2014-05-09
Imports are only added to one addressbook


ostami:
I exported several vcard files from my Android. These were backups. I created new addressbooks call Android-backup-01 and Android-backup-02 in owncloud 6. I then imported each vcard file, selecting Android-backup-01 for my oldest backup and Android-backup-02 for my next to last backup. I'm using firefox 28 from Ubuntu repositories. All the contacts have been imported in only the Android-backup-01 addressbook. No contacts are in the Android-backup-02.

babelouest:
this was fixed in #736 , closing it

PR #735, opened 2014-12-08
fix bug #488

babelouest:
to be able to import in different addressbooks without having to reload the page

### DIFF
--- a/js/addressbooks.js
+++ b/js/addressbooks.js
@@ -456,7 +456,7 @@ OC.Contacts = OC.Contacts || {};
 						backend: $import_into.find('option:selected').attr('backend')
 					}
 				);
-				me.$importFileInput.fileupload('option', 'url', url);
+				me.$importFileInput.fileupload().fileupload('option', 'url', url);
 				me.$importFileInput.attr('disabled', false);
 			} else {
 				me.$importFileInput.attr('disabled', true);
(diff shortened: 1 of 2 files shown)


babelouest:
@jbtbnl , this is another bug that is related to import as well. I made another PR for the stable7 branch (#736) because master and stable7 branches are quite different now in this part of the application.

jbtbnl:
@babelouest I did a quick search to see what the function variable `getStatus` is used for, but I can not find any occurrences of it, except for the definition. So I assume we can delete https://github.com/owncloud/contacts/blob/import-different-addressbooks/js/addressbooks.js#L537

`interval` variable is unused as well:
https://github.com/owncloud/contacts/blob/import-different-addressbooks/js/addressbooks.js#L529

babelouest:
That's possible, I'll test the suppression when I'll have a minute, thanks :-)

Although, one thing I have in my TODO list is to completely rewrite the javascript import UI part.
When I implemented the multi-format, I adapted the existing import UI which already was an modification of a previous import UI, so the technical debt is huge and makes bugs like this one difficult to find and to clean.

babelouest:
@jbtbnl , in the stable branch, the getStatus function is used in an interval: https://github.com/owncloud/contacts/blob/import-different-addressbooks-stable/js/addressbooks.js#L538
That's probably an unfinished cleanup in the master branch.

Anyway, I think this would be better to make another PR for the getStatus thing, I'd like to merge this PR and the stable one too for the next release, this bug is here for too long...

jbtbnl:
@babelouest agreed, we need to push fixes before oC 8 :+1:

babelouest:
:+1: 
I understand that the stable branch is deprecated since oc8 is in alpha mode and based on the master branch, so no need for #736, right ?

jbtbnl:
@babelouest it is up to us and the community how long we want to push fixes to stable7.

babelouest:
@jbtbnl , you're right, no need to remove this PR.

PR #764, opened 2014-12-30
require OC 7.8 not 8, or we don't work with current master

AdamWill:
core has its version set to x.8.y before pre-releases start shipping, then x.9.y when pre-releases are shipping; requiring 8 means the app is disabled until much later.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -4,7 +4,7 @@
 	<name>Contacts</name>
 	<licence>AGPL</licence>
 	<author>Jakob Sack,Thomas Tanghus</author>
-	<require>8</require>
+	<require>7.8</require>
 	<shipped>true</shipped>
 	<description>Address book with CardDAV support.</description>
 	<standalone/>


LukasReschke:
:+1:

AdamWill:
issue not in this change.

AdamWill:
I saw the same issue in a core app and filed a PR for that - https://github.com/owncloud/core/pull/13057 . seems folks decided to just bump the core version to 8.0.0.0, so this probably doesn't need changing now.

DeepDiver1975:
Hmm - we should really only require major owncloud version (like 7,8,9 ...)

jancborchardt:
So what’s the call on this Pull Request? cc @DeepDiver1975 @tomneedham @babelouest?

AdamWill:
IIRC it was kinda made obsolete; I reported a similar issue in core - https://github.com/owncloud/core/pull/13057 - and instead of changing the app requirements, the core versioning scheme was changed instead, which would also have fixed this case.

jancborchardt:
Ok, good call. :) @AdamWill you are very welcome in our IRC channel #owncloud-contacts

Issue #771, opened 2015-01-11
File selector in contacts import uses wrong file type filter


Pastafarianist:
I am using ownCloud v7.0.3.

In "contacts" app, there is an option to select the format of import. After choosing "Gmail CSV" and opening the file selector, the file filter is set up to allow only for vCard files, not CSV.

jbtbnl:
@Pastafarianist have you tested this against the development branch? The import dialog has been redesigned, and possibly fixes your issue.

babelouest:
He has a point, in the stable branch, there is still only vcard files as input filter: https://github.com/owncloud/contacts/blob/stable7/templates/contacts.php#L54
=> https://github.com/owncloud/contacts/pull/776
The master branch is clean though

babelouest:
Is the problem still there ? A related PR has been merged a few months ago #795

Pastafarianist:
In ownCloud 8.2.2 the filter is disabled entirely.

PR #776, opened 2015-01-15
fix import fileupload filter which had only text/vcard and text/x-vcard (stable7 backport)

babelouest:


jbtbnl:
@Pastafarianist please check if this fix solves #771 for you!

Pastafarianist:
@jbtbnl I'm on stable branch and I'm afraid I can't do that.

dniku:
@jbtbnl Yes, it is fixed. Thanks!

jancborchardt:
@babelouest @DeepDiver1975 is this still current? Do we do stable7 backports?

DeepDiver1975:
I vote for looking at the future of this app in the context of oc8 and oc8.1

PR #800, opened 2015-02-06
[WIP] Fix unit tests

LEDfan:
......

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: php
 php:
-  - 5.3
   - 5.4
   - 5.5
 


LEDfan:
```
PHP Fatal error:  Call to a member function getUId() on a non-object in /var/www/owncloud/lib/private/tagmanager.php on line 79
```

LEDfan:
cc @LukasReschke @DeepDiver1975 (you wrote the unit test)
This is what causing the unit tests to fail:
This line https://github.com/owncloud/contacts/blob/master//tests/lib/addressbookprovider_test.php#L61-61 calls the delete function. This calls a hook here: https://github.com/owncloud/contacts/blob/master/lib/hooks.php#L76-L77 . Which tries to call the tagmanager. Of course the tagmanager won't work: https://github.com/owncloud/core/blob/master/lib/private/tagmanager.php#L78-L82 , because we aren't logged in......... 

I'll mock the tagmanager.

DeepDiver1975:
superseeded by #891

PR #816, opened 2015-02-10
Fix 807 810

babelouest:
fixes small problems with import that were not detected before for some reason.

### DIFF
--- a/formats/import_csv_gmail_connector.xml
+++ b/formats/import_csv_gmail_connector.xml
@@ -162,12 +162,12 @@ author: Nicolas Mora mail@babelouest.org
 	</import_entry>
 	
 	<import_entry name="Phone 1 - Type" enabled="true" remove="*">
-		<vcard_entry property="PHONE" group="phone1" value="type">
+		<vcard_entry property="TEL" group="phone1" value="type">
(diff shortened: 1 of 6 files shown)


gstuartj:
PR #816 is importing the home phone number field now, but it still seems to be missing mobile phone numbers. Also, if contact entries have multiple associated home phone numbers they are listed in one field separated by the string " ::: " rather than a comma like other import formats.

Address fields are also not parsed correctly and tend to have a lot of ", ::: ," delimiters sprinkled in.

babelouest:
@gstuartj , it looks like other problems then, can you post new issues, one per problem so they can be better addressed ? If you comment this PR, the probability they are lost in the conversation are high...

gstuartj:
Done.

gstuartj:
#818 fixed by 8f29f21

DeepDiver1975:
let me bring back ci - see #891 - then we add proper unit tests to this one an merge it

DeepDiver1975:
@babelouest @gstuartj Any chance to get some sample file to be used as basis to build unit test? THX

DeepDiver1975:
lets move this in :+1:

PR #824, opened 2015-02-13
Avoid "Undefined offset: 0"

martin-rueegg:
$addressBook is always an array. either empty or having one entry.

owncloud#740
nerzhul/ocsms#39

### DIFF
--- a/lib/backend/shared.php
+++ b/lib/backend/shared.php
@@ -87,8 +87,22 @@ class Shared extends Database {
 
 		// Not sure if I'm doing it wrongly, or if its supposed to return
 		// the info in an array?
-		$addressBook = (isset($addressBook['permissions']) ? $addressBook : $addressBook[0]);
-
+		// https://github.com/owncloud/core/blob/master/lib/private/share/share.php
+		// \-> getItemSharedWithBySource()
+		//     \-> self::getItems()
+		//          -> return array();
+		//       or -> return formatResult(); 
+		//             \-> https://github.com/owncloud/contacts/blob/master/lib/share/addressbook.php
+		//                 \-> formatItems()
+		//                      \--> which in turn also always returns an array
+		// thus $addressBook is an array. either empty or having one entry
+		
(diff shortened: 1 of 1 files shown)


martin-rueegg:
@enoch85, if you could install this version for testing, it would be great. same applies to @LEDfan.

martin-rueegg:
not sure about the [inspection](https://scrutinizer-ci.com/g/owncloud/contacts/inspections/b77826ab-c3c0-47ed-9d0b-10469018aafd) and the [branch-test](https://travis-ci.org/owncloud/contacts/builds/50612607). it seems to me that only the [second commit](https://github.com/martin-rueegg/contacts/commit/f941a3083877b3ea1918083f4a5ddcde313a2059) has been checked and failed, since i just removed the comments. the [first commit](https://github.com/martin-rueegg/contacts/commit/95cdb64b143dd2851009af5754dc0fcea82e9d1c) contains the real change.

LEDfan:
I'll give this a test today, and let know if it works.

deMattin:
For me the following works well:

Replacing in /owncloud/apps/contacts/lib/backend/shared.php:
Line 88-90:

> ```
>   // Not sure if I'm doing it wrongly, or if its supposed to return
>   // the info in an array?
>   $addressBook = (isset($addressBook['permissions']) ? $addressBook : $addressBook[0]);
> ```

With:

> ```
>   if(count($addressBook) == 0) {
>       return null;
>   }
>   $addressBook = $addressBook[0];
> ```

I use an external app, that uses the contacts app (rainloop)
Without the patch, rainloop is able to get mailadresses from contacts but I get the Error:
Error   PHP Undefined offset: 0 at /var/www/owncloud/apps/contacts/lib/backend/shared.php#90

With the fix, the error doesn't occur any more and the contacts access through rainloop works fine as well.

Please add this in ownCloud master.

LEDfan:
Fix works. Thanks!

PR #835, opened 2015-02-25
Correctly only check stable* branches and master and ignore php 5.3

nickvergessen:
@DeepDiver1975


nickvergessen:
ermm _hust_

DeepDiver1975:
:+1:

PR #838, opened 2015-02-27
Show current age instead of year of original birthday

xn--nding-jua:
In current version the year of the original birthday is shown. But more interesting is the age of the person in the current year. So the changed code will present the age instead of the year.

### DIFF
--- a/lib/contact.php
+++ b/lib/contact.php
@@ -828,7 +828,10 @@ class Contact extends VObject\VCard implements IPIMObject {
 			$vevent->add('DURATION', 'P1D');
 			$vevent->{'UID'} = $this->UID;
 			$vevent->{'RRULE'} = 'FREQ=YEARLY';
-			$vevent->{'SUMMARY'} = $title . ' (' . $date->format('Y') . ')';
+			$today = New \DateTime(date('d.m.Y'));
+			$bdaythisyear = New \DateTime($date->format('d') . '.' . $date->format('m') . '.' . $today->format('Y'));
+			$age = $bdaythisyear->diff($date);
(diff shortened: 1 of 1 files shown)


jakobsack:
:-1: I recommend to ignore this pull request!

With this patch you do not only update a recurring event on the first day of each year, but it might give you wrong results, too. For example, if it is December and you are planning something for January all birthdays in January will show the wrong age (n-1).

xn--nding-jua:
Hey Jakob,
generally a birthday lies in the past, so the calculation should be right for most use cases. We could test the calculated age to be >=0 and <120. Otherwise the absolute date could be used again?

jakobsack:
In my opinion the calculated age should be correct in any use case. And at least in the implementation you suggested this is not the case. If you open the calendar for any date that is not in the current year it will show the same age for last year, this year (correct) and next year.

xn--nding-jua:
Hey, you are right - the calculated age is based on the todays date and not the current date the calendar is showing. But is there a possibility to get the displayed date of the calendar? I fear no.

If yes, only the line "$today = New \DateTime(date('d.m.Y'));" has to be replaced by the date of the current calendar-event showing.

As my php-experience is limited my contribution has to end here. Maybe someone other can continue with this or the pull-request really has to be rejected.

jakobsack:
I'm afraid that this is a limitation of the ical standard we can't do anything about. ownCloud will never know what date I'm looking at on my phone or my desktop calendar app :-/ And that's how far these calendar entries go.

I do not want to scare you away and I hope that you find a way to contribute to ownCloud!

PR #877, opened 2015-04-11
fix Undefined index error ehen updatein thumbnail of contacts

### LINKED ISSUE #767: OC8: Updating contacts fails
Version: Owncloud 8 alpha 1

Reproduce: Change contacts thumbnail on a mobile device (android with sync via davdroid)

Error: Operation fails, error is attached (in operation cacheThumbnail)

Log: 

"app":"PHP","message":"Undefined index: remove at \/var\/www\/owncloud\/apps\/contacts\/lib\/utils\/properties.php#303","level":3,"time":"January 04, 2015 17:56:40"}

derkostka:
fixes #767

### DIFF
--- a/lib/utils/properties.php
+++ b/lib/utils/properties.php
@@ -300,7 +300,7 @@ Class Properties {
 				return $cache->get($key);
 			}
 		} else {
-			if ($options['remove'] === false && $options['update'] === false){
+			if (isset($options['remove']) === false && isset($options['update']) === false){
 				return $cache->get($key);
 			}
 		}
--- a/lib/utils/properties.php
+++ b/lib/utils/properties.php
@@ -300,11 +300,11 @@ Class Properties {
 				return $cache->get($key);
 			}
 		} else {
-			if (isset($options['remove']) === false && isset($options['update']) === false){
+			if ( (isset($options['remove']) && $options['remove'] === false) 
+			&&   (isset($options['update']) && $options['update'] === false) ){
 				return $cache->get($key);
 			}
 		}
-
 
 		if (isset($options['remove']) && $options['remove']) {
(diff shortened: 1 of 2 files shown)


derkostka:
This commit is for free software in the public domain.
@PVince81 : Please backport to OC 8.0.3 and also OC 7. I´ve had this error since OC 7 and updating contacts thumbnails was not possible because of this :+1:

lib/utils/properties.php
```diff
@@ -300,7 +300,7 @@ public static function cacheThumbnail($backendName, $addressBookId, $contactId,
 				return $cache->get($key);
 			}
 		} else {
-			if ($options['remove'] === false && $options['update'] === false){
+			if (isset($options['remove']) === false && isset($options['update']) === false){
```
LEDfan: This will fix the issue, but it won't have the same result.

LEDfan: I.e. the when 'remove' property exists but is false this code will not be called. Which should be the case AFAIK.

derkostka:
@LEDfan  tbh, i do not understand what the else statement is for ?

What about this, we check with "isset" if the key is not null, and if it is, we check if it is set 'false'
            if ( (isset($options['remove']) && $options['remove'] === false) 
            &&   (isset($options['update']) && $options['update'] === false) ){
                return $cache->get($key);
            }

?

derkostka:
any idea on this ? if not, let´s close this one also, but the issue remains for me ...

PVince81:
I haven't tested but your code change looks good :+1: 
I might have missed that `$vevent` part when doing the Sabre 2.1 adaptation.

Is it broken on 8.0 and 7 too ?

PR #887, opened 2015-04-27
Make birthday events non-blocking

jakobsack:
Currently birthdays are blocking an entire day on calendars.

This patch utilizes the "TRANSP" property of RFC 2445 (4.8.2.7 Time Transparency) to change this behaviour.

Change works as expected on Windows Phone 8.1, no conflicts with other events on the same day are shown.


derkostka:
Nice addition ! +1

PR #889, opened 2015-04-27
fix #888 - $vevent was not initialized 

derkostka:
This commit is under the MIT license, you can use it for any purpose, of course !

PR #895, opened 2015-05-08
Fix creation of ics files

jakobsack:
There was a wrong variable name in contacts.php

Please test this PR as I can not do this at the moment. The ics file should contain one line similar to this:

```
DTSTART;VALUE=DATE:20070628
```

For more information on that matter search RFC 5545 for "VALUE=DATE"

### DIFF
--- a/lib/contact.php
+++ b/lib/contact.php
@@ -832,7 +832,7 @@ class Contact extends VObject\VCard implements IPIMObject {
 			$vevent->DTSTART->setDateTime(
 				$date
 			);
-			$event->DTSTART['VALUE'] = 'date';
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
this is already fixed by another pr afaik

jakobsack:
Can't find any PR touching this line. Closest match is #889 which does ignore this occurance of $event. But maybe I overlooked something.

DeepDiver1975:
you are right - https://github.com/owncloud/contacts/pull/889#issuecomment-100038482

DeepDiver1975:
:+1:

Issue #888, opened 2015-04-27
DAVdroid sync is broken on Master - $vevent is not defined


derkostka:
Owncloud 8.1.alpha2 and Contacts-Master

I have exported all contacts from owncloud and imported them again, which worked well. Then i deleted all contacts on my Andorid device, they get synced 100% and then deleted again. The log is:

Log:
{"reqId":"VT6VXn8AAAEAAB0Msm4AAAAO","remoteAddr":"xxx","app":"PHP","message":"Undefined variable: vevent at \/var\/www\/owncloud\/apps2\/contacts\/lib\/contact.php#831","level":3,"time":"April 27, 2015 22:00:31"}
{"reqId":"VT6VXn8AAAEAAB0Msm4AAAAO","remoteAddr":"xxx","app":"PHP","message":"Call to a member function add() on a non-object at \/var\/www\/owncloud\/apps2\/contacts\/lib\/contact.php#831","level":3,"time":"April 27, 2015 22:00:31"}

Is this a DAVdroid issue or can somebody fix it here ? With OC 8.0 everything was perfect.

derkostka:
I guess we need something like:
$vevent = new \Sabre\VObject\Component\VEvent();
within contact.php right ?

derkostka:
the error at least is fixed with #889

derkostka:
New request #896

derkostka:
still and even with that, sync does not work for me. please check before 8.1 is released.

jakobsack:
merged #896

PR #896, opened 2015-05-10
rebased fix for #888

derkostka:
removed unneeded line in getBirthdayEvent


derkostka:
This change is under MIT Licence. You can do what you want with it :-)
please check @jakobsack @DeepDiver1975

PVince81:
Makes sense :+1:

jakobsack:
looks good :+1:

PR #901, opened 2015-05-14
Added german entry names for outlook-csv importer

MatthiasMMayr:
As of csv field names beeing localized, the importer was only abled to import csv files from english and french versions of outlook.  
With this little change, also german versions will be supported. Source for my altnames was an dummy export from outlook 2013. Tested with Owncloud 8.0.


babelouest:
Sehr gut!

DeepDiver1975:
:+1:

Issue #880, opened 2015-04-15
Import fails: JSON.parse: unexpected character


jancborchardt:
When trying to import a Contacts.vcf I just exported from my ownCloud instance, I get this yellow notification error:
`SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data: parsererror`

In the JS console:
```
"Import clicked" addressbooks.js:379:3
Object { readyState: 4, getResponseHeader: .ajax/C.getResponseHeader(), getAllResponseHeaders: .ajax/C.getAllResponseHeaders(), setRequestHeader: .ajax/C.setRequestHeader(), overrideMimeType: .ajax/C.overrideMimeType(), statusCode: .ajax/C.statusCode(), abort: .ajax/C.abort(), state: .Deferred/r.state(), always: .Deferred/r.always(), then: .Deferred/r.then(), 12 de plus… } storage.js:707:4
"jqXHR" Object { readyState: 4, getResponseHeader: .ajax/C.getResponseHeader(), getAllResponseHeaders: .ajax/C.getAllResponseHeaders(), setRequestHeader: .ajax/C.setRequestHeader(), overrideMimeType: .ajax/C.overrideMimeType(), statusCode: .ajax/C.statusCode(), abort: .ajax/C.abort(), state: .Deferred/r.state(), always: .Deferred/r.always(), then: .Deferred/r.then(), 12 de plus… } storage.js:12:2
"fail" Object { disabled: false, create: null, dropZone: Object, replaceFileInput: true, singleFileUploads: true, limitMultiFileUploadSizeOverhead: 512, sequentialUploads: false, forceIframeTransport: false, multipart: true, recalculateProgress: true, 46 de plus… }
```

And this in the owncloud.log
```
{"reqId":"ATvfOgsJxMq6L+3KZRJO","remoteAddr":"127.0.0.1","app":"no app in context",
"message":"Accessing the API class is deprecated! Use the appropriate services instead!",
"level":0,"time":"2015-04-15T20:20:42+00:00","method":"GET",
"url":"\/owncloud\/index.php\/apps\/contacts\/addressbooks\/"}
{"reqId":"jyp57Ts+Be+S1lEiMAAU","remoteAddr":"127.0.0.1","app":"no app in context",
"message":"Accessing the API class is deprecated! Use the appropriate services instead!",
"level":0,"time":"2015-04-15T20:20:48+00:00","method":"POST",
[...]

PR #903, opened 2015-05-19
fix bug #880

babelouest:
there was no $root parameter passed to the VObject/Parameter creation

### DIFF
--- a/lib/connector/importconnector.php
+++ b/lib/connector/importconnector.php
@@ -62,11 +62,11 @@ abstract class ImportConnector {
 		if (isset($property) && isset($importEntry) && isset($value)) {
 			if (isset($importEntry->vcard_entry)) {
 				if (isset($importEntry->vcard_entry['type'])) {
-					$property->parameters[] = new \Sabre\VObject\Parameter('TYPE', ''.StringUtil::convertToUTF8($importEntry->vcard_entry['type']));
+          $property->add('TYPE', StringUtil::convertToUTF8($importEntry->vcard_entry['type']));
 				}
 				if (isset($importEntry->vcard_entry->additional_property)) {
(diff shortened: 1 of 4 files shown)


lib/connector/importconnector.php
```diff
@@ -66,7 +66,7 @@ protected function updateProperty(&$property, $importEntry, $value) {
 				}
 				if (isset($importEntry->vcard_entry->additional_property)) {
 					foreach ($importEntry->vcard_entry->additional_property as $additionalProperty) {
-						$property->parameters[] = new \Sabre\VObject\Parameter(''.$additionalProperty['name'], ''.$additionalProperty['value']);
+						$property->parameters[] = new \Sabre\VObject\Parameter($root, ''.$additionalProperty['name'], ''.$additionalProperty['value']);
```
evert: I'm surprised this worked in the first place, because the constructor signature actually looks like this:

```
public function __construct(Document $root, $name, $value = null) {
```

regardless it's probably better to change this line entirely from:

```
$property->parameters[] = new \Sabre\VObject\Parameter($root, ''.$additionalProperty['name'], ''.$additionalProperty['value']);
```

to:

```
$property->add($additionalProperty['name'], $additionalProperty['value']);
```

The current way is not recommended because it will break when working with multiple parameters with the same name.

jancborchardt:
Looks good. #880 was fixed meanwhile for me by something else it seems, but this also works nicely.

lib/connector/importconnector.php
babelouest: Thanks for the feedback, I will change it for all $property->parameters[] calls too in the same time

babelouest:
I recreated a pull request to fix this issue and #905 at the same time: #906 , so this PR becomes deprecated.
@jancborchardt , can you take a look ?

PR #910, opened 2015-05-27
make search case insensitive

### LINKED ISSUE #813: Search is case sensitive
Owncloud 8.0 (8.0.0-5 package, Ubuntu, fresh install)
PHP version: PHP 5.5.9-1ubuntu4.5

When searching contact entries the search operation for all fields (and importantly, the name field) is case sensitive. This behavior is probably not expected by the user; it definitely caused me some confusion.

ChristophWurst:
I hope this doesn't break anything else :smile: 

fixes #813, https://github.com/owncloud/mail/issues/197

### DIFF
--- a/lib/addressbookprovider.php
+++ b/lib/addressbookprovider.php
@@ -121,8 +121,8 @@ SQL;
 		$params = array();
 		foreach ($searchProperties as $property) {
 			$params[] = $property;
-			$params[] = '%' . $pattern . '%';
-			$query .= '(`name` = ? AND `value` LIKE ?) OR ';
+			$params[] = '%' . strtolower($pattern) . '%';
+			$query .= '(`name` = ? AND LOWER(`value`) LIKE ?) OR ';
 		}
 		$query = substr($query, 0, strlen($query) - 4);
 		$query .= ')';


jancborchardt:
@zinks- please review. :)

jancborchardt:
Also @gstuartj @jelle-e @ntassani @babelouest please review as this fixes https://github.com/owncloud/contacts/issues/813

babelouest:
Looks good, it fixes the bug for me

jancborchardt:
@gstuartj @jelle-e @ntassani @zinks- another review please! :)

@wurstchristoph can you look into the Travis failure?

jelle-e:
Totally fixed! Also when using the API to search.
Thanks.

jelle-e:
Note: I just heard 'strtolower' function does not work correctly with utf 8 strings.

ChristophWurst:
@jelle-e hm, should we use the SQL `LOWER` function instead?

`$query .= '(`name` = ? AND LOWER(`value`) LIKE LOWER(?)) OR ';`

jelle-e:
@wurstchristoph or maybe mb_strtolower? I'm not into coding that much, so I can't help you as such.

From the manual: http://php.net/manual/en/function.mb-strtolower.php
By contrast to strtolower(), 'alphabetic' is determined by the Unicode character properties. Thus the behaviour of this function is not affected by locale settings and it can convert any characters that have 'alphabetic' property, such as A-umlaut (Ä).

jancborchardt:
What’s the status on this btw? Mergable? :)

Georg-Git:
I found a simpler way without mb_strtolower:
After changing 
"LIKE"  to 
"COLLATE utf8_general_ci LIKE"
searching is case insensitive.

Old Query:

``` SQL
$query .= '(`name` = ? AND `value` LIKE ?) OR ';
```

New Query:

``` SQL
$query .= '(`name` = ? AND `value`COLLATE utf8_general_ci LIKE ?) OR ';
```

@ChristophWurst  : what do think?

DeepDiver1975:
ownClouds database kung-fu support ILIKE - we shall use that one.

DeepDiver1975:
I'll submit a PR using ILIKE

ChristophWurst:
> I'll submit a PR using ILIKE 
> awesome, thanks!

PR #906, opened 2015-05-23
Fix #880 #905

babelouest:
bugs with multiple values and parameters on import

PR #915, opened 2015-06-06
Fix standard vcard import

### LINKED ISSUE #566: address book import does not work properly
First, it's not consistent how the import is solved in the UI:

Address Book:
1) Import via "Contacts" app > Settings > Import
2) Import via a previously uploaded VCF file in the "Files"-App

Calendar:
Import only via click on a previously uploaded ICS file in the "Files"-App

Second, if importing via "Files"-App the progress bar is broken:

Address Book:
After choosing the Address Book to import I see the progress bar "Importing..." until I reload the page. The progress bar is animated but there is no progress and no "Finished" message. If I reload the page after some time no contacts were added.

BernhardPosselt:
Fixes #566 in a way that that if you select standard vcard format the import works. This feature is crucial because if you initially start syncing most mobile OSes (looking at you iOS and Android) make it an absolute pain to copy contacts into a new group. The only way to do this is by exporting your contacts as vcard and importing them again.

This fix is not best practice, but it gets stuff working again. And lets be honest, almost the entirety of the controller needs to be refactored which takes too long to fix this simple issue. I'd say lets apply this fix and wait for some maintainer to rewrite huge portions of the app. Or simply kill the app if a better and maintained alternative is already around

@jancborchardt @LukasReschke @DeepDiver1975 @MorrisJobke @halfline

### DIFF
--- a/lib/controller/importcontroller.php
+++ b/lib/controller/importcontroller.php
@@ -125,7 +125,7 @@ class ImportController extends Controller {
 			$view->mkdir('imports');
 		}
 
-		$content = \OC_Filesystem::file_get_contents($path . '/' . $filename);
+		$content = \OC\Files\Filesystem::file_get_contents($path . '/' . $filename);
 		if($view->file_put_contents('/imports/' . $filename, $content)) {
 			$progresskey = 'contacts-import-' . rand();
 			$response->setParams(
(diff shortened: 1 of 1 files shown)


jancborchardt:
@babelouest how does this relate to https://github.com/owncloud/contacts/pull/906 ?

@Raydiation always please ask the people to review who opened the issue and discussed in there. In this case: @tbeitter @imago-maeror @domcars0

babelouest:
@jancborchardt , afaik, this doesn't fix #906 but only #566, which is good too

DeepDiver1975:
:+1:

PR #921, opened 2015-06-14
Fixed error handling in getContact if  $id not an array.

schuemann:
I had some group<->contact associations in the oc_vcategory_to_object table for contacts that where no longer existing. The cause of these is another topic, but  it resulted in a lot of errors like this in the log, when I accessed the contacts app on the webpage:

> implode(): Invalid arguments passed at [...]/contacts/lib/backend/shared.php#139

I made the implode optional (only if $id is an array) to fix this error.


DeepDiver1975:
looks good - thx :+1:

PR #924, opened 2015-06-17
Federated cloud

schiessle:
allow user to add Federated Cloud IDs to the address book 

cc @jancborchardt @tomneedham

### DIFF
--- a/lib/utils/properties.php
+++ b/lib/utils/properties.php
@@ -48,7 +48,7 @@ Class Properties {
 	 *
 	 * @var array
 	 */
-	public static $multiProperties = array('EMAIL', 'TEL', 'IMPP', 'ADR', 'URL');
+	public static $multiProperties = array('EMAIL', 'TEL', 'IMPP', 'ADR', 'URL', 'CLOUD');
 
 	/**
 	 * Properties to index.
@@ -57,7 +57,7 @@ Class Properties {
 	 */
 	public static $indexProperties = array(
 		'BDAY', 'UID', 'N', 'FN', 'TITLE', 'ROLE', 'NOTE', 'NICKNAME',
-		'ORG', 'CATEGORIES', 'EMAIL', 'TEL', 'IMPP', 'ADR', 'URL', 'GEO');
+		'ORG', 'CATEGORIES', 'EMAIL', 'TEL', 'IMPP', 'ADR', 'URL', 'GEO', 'CLOUD');
 
 	/**
 	 * Get options for IMPP properties
@@ -182,6 +182,12 @@ Class Properties {
 					'WORK' => (string)$l->t('Work'),
 					'HOME' => (string)$l->t('Home'),
 					'INTERNET' => (string)$l->t('Internet'),
+					'OTHER' =>  (string)$l->t('Other'),
+				);
+			case 'CLOUD':
+				return array(
+					'HOME' => (string)$l->t('Home'),
+					'WORK' => (string)$l->t('Work'),
 					'OTHER' =>  (string)$l->t('Other'),
 				);
 		}
--- a/templates/contacts.php
+++ b/templates/contacts.php
@@ -359,6 +359,11 @@ vendor_style('contacts', array(
 			</ul>
 		</li>
 		<li>
+			<ul class="cloud propertylist hidden">
+				<h3><?php p($l->t('Federated Cloud ID')); ?></h3>
+			</ul>
+		</li>
+		<li>
 			<ul class="impp propertylist hidden">
 				<h3><?php p($l->t('Instant messaging')); ?></h3>
 			</ul>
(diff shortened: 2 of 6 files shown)


templates/contacts.php
```diff
@@ -359,6 +359,11 @@ class="coords"
 			</ul>
 		</li>
 		<li>
+			<ul class="cloud propertylist hidden">
+				<h3><?php p($l->t('Federated Cloud Id')); ?></h3>
```
jancborchardt: »Federated Cloud ID« please – ID both letters uppercase. ;)

jancborchardt: »Federated Cloud ID« please – ID both letters uppercase. ;)

templates/contacts.php
```diff
@@ -419,6 +425,20 @@ class="coords"
...
 	</div>
+	<div class="cloud" type="text/template">
+		<li data-element="cloud" data-checksum="{checksum}" class="propertycontainer">
+			<span class="parameters">
+				<input type="checkbox" class="parameter tooltipped rightwards" data-parameter="TYPE" name="parameters[TYPE][]" value="PREF" title="<?php p($l->t('Preferred')); ?>" />
+				<select class="rtl type parameter" data-parameter="TYPE" name="parameters[TYPE][]">
+					<?php print_unescaped(OCP\html_select_options($_['cloudTypes'], array())) ?>
+				</select>
+			</span>
+			<input type="text" class="nonempty value" name="value" value="{value}" x-moz-errormessage="<?php p($l->t('Please specify a valid federated cloud id.')); ?>" placeholder="<?php p($l->t('someone@owncloud.example.com')); ?>" required />
```
jancborchardt: Same here: »Federated Cloud ID«. Uppercase Federated, Cloud and ID.

Also the example format should be the same as in the ownCloud sharing dialog: `username@example.com/owncloud`

templates/contacts.php
```diff
@@ -419,6 +425,20 @@ class="coords"
...
+			<span class="listactions">
+				<a role="button" class="icon-delete action delete tooltipped leftwards" title="<?php p($l->t('Delete federated cloud id')); ?>"></a>
```
jancborchardt: And same here: »Delete Federated Cloud ID«

jancborchardt:
`Federated Cloud ID` is the only way to write it. ;) Not ~~Federated cloud id~~ or ~~Federated Cloud Id~~ or the like.

schiessle:
@jancborchardt all your comments are addressed

jancborchardt:
Can you review this @babelouest @LEDfan @jbtbnl @derkostka? :) Also – everyone join our IRC channel #owncloud-contacts please!

derkostka:
Ok seems good to me and work (tested adding a new user, then download the vcard and opened in a text editor). Just curious: What happens if there is an @ in the username ? do i add the link like that: user@someuniversityasloginemail@somecloudserver.com/owncloud
:+1:

schiessle:
@derkostka we should split uid and server from the right side. But good question, I will double-check this in core

schiessle:
@derkostka really good catch, indeed this was not handled correctly in core. Here is the fix for core https://github.com/owncloud/core/pull/17006

schiessle:
would be great to have a second reviewer, thanks!

schiessle:
@aidanamavi can you please review this PR? Thanks a lot!

jancborchardt:
Btw @derkostka it would be really cool if you join our IRC channel #owncloud-contacts :)

lib/utils/properties.php
```diff
@@ -184,6 +184,12 @@ public static function getTypesForProperty($prop) {
 					'INTERNET' => (string)$l->t('Internet'),
 					'OTHER' =>  (string)$l->t('Other'),
 				);
+			case 'CLOUD':
+				return array(
+					'WORK' => (string)$l->t('Work'),
```
jancborchardt: This should be sorted so that »Home« is first, and »Work« is second. Also see https://github.com/owncloud/contacts/issues/931 about that.

templates/contacts.php
```diff
@@ -384,6 +389,7 @@ class="coords"
 			<option value="ADR"><?php p($l->t('Address')); ?></option>
 			<option value="NOTE"><?php p($l->t('Note')); ?></option>
 			<option value="URL"><?php p($l->t('Web site')); ?></option>
+			<option value="CLOUD"><?php p($l->t('Federated Cloud')); ?></option>
```
jancborchardt: Should also say »Federated Cloud ID« here, right?

jancborchardt:
Except the 2 notes above it looks good! :)

templates/contacts.php
schiessle: but we also use "Email" and not "Email address" and "Phone" and not "Phone number"... That was the rational behind my decision on calling it just "Federated Cloud".... But I don't have a strong opinion here, I can also change it.

schiessle:
@jancborchardt fixed :smile:

templates/contacts.php
jancborchardt: Good point, yeah.

jancborchardt:
:+1: from my POV. Let’s get this in.

PR #926, opened 2015-06-19
fix app loading with avatars disabled

mroi:
The contacts app fails to load with JavaScript errors, when avatars are disabled. I am not sure this is the proper fix, but as a workaround, it works for me.


DeepDiver1975:
tested - works - thx a lot :+1:

Issue #931, opened 2015-06-23
Email, Address and any new field is »Work« by default


jancborchardt:
Instead the default should be »Home«. Which we should also rename to »Personal«.

@babelouest @derkostka @schiesbn @ChristophWurst how can we change this?

schiesbn:
This part was easy, see the pull request above. But I have another problem: I can't change the type. If I click on the "Work" label (or "Home" with the pr above) nothing happens. I would expect a drop-down to chose the type. Until now I couldn't find the right place to fix it... any pointers or does someone takes care about it?

jancborchardt:
For me, the dropdown shows up. Any errors or JS issues? Does it not work in different browsers either?

schiesbn:
Hm, on Chromium it works for me too, but not on Iceweasel 38.0.1... No errors on the JS console

schiesbn:
Another thing... sometimes I see a "PREF" as type. But it is not really reproducible most of the time a reload already help.s

schiesbn:
@nickvergessen could reproduce the problem with firefox. We discovered that the drop down gets opened, but in the background. If you clicked on the label you can navigate through the options with "up", "down"... Probably a z-index problem. @jancborchardt do you think you can have a look at it?

PR #934, opened 2015-06-25
default should always be home

schiessle:
default type should always be home

fix https://github.com/owncloud/contacts/issues/931


DeepDiver1975:
:+1:

PR #963, opened 2015-07-02
Translate the noneSelectedText on the groups multiselect

### LINKED ISSUE #960: 8.1rc2 not translated lang on Groups selectbox
TestEnviro:
8.1rc2

Step to reproduce:
1. create a new contact
2. watch on the group selectbox text "select options"

Expected result
the "select options" should be translated into choosen lang

Actual result
no translation

DeepDiver1975:
fixes #960


jancborchardt:
Awesome :+1: good detail

PR #965, opened 2015-07-03
Redirect to default page if we are not logged in

DeepDiver1975:
@jancborchardt

### DIFF
--- a/lib/dispatcher.php
+++ b/lib/dispatcher.php
@@ -11,8 +11,8 @@
 namespace OCA\Contacts;
 
 use OCP\AppFramework\App as MainApp;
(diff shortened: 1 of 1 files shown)


jancborchardt:
Works, however the redirect is wrong. Instead of http://127.0.0.1/owncloud/index.php/apps/contacts/#979 it changed to http://127.0.0.1/owncloud/index.php/apps/files/#979 and opened Files.

DeepDiver1975:
hmm - an issue we need to dig deeper - maybe an issue in core? No idea. Let's merge this anyway - this is better then a white-screen-of-death

jancborchardt:
True.

Issue #961, opened 2015-07-02
8.1rc2 missing searchbar on header


libasys:
TestEnviro:
8.1rc2

Step to reproduce:
1. goto contacts
2. watch right side of the header

Expected result
you should see a search bar

Actual result 
no search bar is visible

DeepDiver1975:
I'll bring back search for 8.2

sabl0r:
Still missing in 8.2. Any update on this?

aspdye:
Seems to be not implemented into the new version :( but why?

piflixe:
I am really missing this feature too. Search field is displayed for a second while the site is loading, but then it disappears. I would really appreciate if it is reincluded in an upcoming update. 

Using Owncloud 8.1.3  with contacts app 0.4.0.0

Is the search field Owncloud or Contacts app functionality?

aspdye:
I must honestly admit that I am quite surprised that you can not make it after more than a month that the calendar and contacts a search function (one of the most important functions for a calendar, and contacts) gets back. It is incomprehensible to me that you could publish this update at all (8.1) and it still does not could you fix with 8.2.

That this should only be fixed with 9.0, is not as good as the normal quality of ownCloud, because this will fix this problem first in February of next year. 

The use of 8.0 (or even 7.0) does not constitute a real alternative for some of us ;)

Calendar+ and Contacts+ (with which one after the update you could still search) are not longer in an really active development (Libasys hasn't realeased any updates since over an month).

But I have to say: I am more a man of ideas, support and marketing. Not for Coding. Therefore, it is difficult for me to assess the situation with calendar and contacts.

But I think we can not change anything here. Calendar and contacts are community projects in principle.

jancborchardt:
@aspdye we know it’s unfortunate for the time being, but as you say Calendar and Contacts are community projects. There will already be massive improvements with the dav sync in core to make sure that’s stable. And then the app frontends themselves will also be updated.

aspdye:
This Time the calendar gets refactored. Is the Same planend for Contacts Too?

jancborchardt:
@aspdye yes, see https://github.com/Henni/contactsrework by @Henni @irgendwie etc. – help is very welcome. :)
This is because the codebase is quite old and with sync in core we only need the frontend. Hence it’s better to start with a new approach. @babelouest @jbtbnl would be cool if you join again. :)

Issue #863, opened 2015-03-26
Search Field in Contacts does not longer work


blainy:
Since i have updated ownCloud from 8.0 to 8.0.2 the Search Field in Contacts does not longer work. It just happens nothing when i enter some Name i am searching for.

cova-it:
+1

hope55:
The search results will be displayed at the very end of the page, but I don't want to scroll down my 1000 contacts to see the results. The contact list sould be filtered instead. The old dropdown list under the search field was still better than this one.

rvega:
+1.  I couldn't find the search results on my 1200+ list until I came here to report a bug and read these comments.

bene33:
+1

rnalrd:
Same here

jancborchardt:
Closing as duplicate of https://github.com/owncloud/contacts/issues/961 – more discussion there, less +1 ;)

Issue #792, opened 2015-01-27
searchbar does not work


timmie:
When inserting a name of the contact in the search bar and hitting enter nothing happens.

OC version: 7

RealRancor:
Hi,

please fill out the issue template provided here:

https://raw.githubusercontent.com/owncloud/core/master/issue_template.md

to provide more informations about your environment and possible log entries which could help debugging this.

PR #966, opened 2015-07-03
Bring back search functionality - fixes #335, #409, #792, #863, #961

DeepDiver1975:
@jancborchardt display of empty results looks odd - any ideas? THX

### DIFF
--- a/css/contacts.css
+++ b/css/contacts.css
@@ -893,3 +893,8 @@ tbody tr.contact.active, tbody tr.contact:hover {
 	#contactsHeader tr td.email { display: none; }
 	#contactsHeader tr td.cloud { display: none; }
 }
+
+#searchresults {
+	margin-top: 0 !important;
+	margin-left: 10px;
+}
--- a/js/search.js
+++ b/js/search.js
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2015
+ *
+ * This file is licensed under the Affero General Public License version 3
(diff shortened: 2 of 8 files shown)


DeepDiver1975:
with some results:
bildschirmfoto von 2015-07-04 00-13-10

no results:
bildschirmfoto von 2015-07-04 00-10-41

jancborchardt:
If there’s no results, then the list with contacts should not be shown. Only the »No search results in other places« #emptycontent thing. Then there’s also no overlap, and clear feedback.

jancborchardt:
(Reviewing now)

DeepDiver1975:
> If there’s no results, then the list with contacts should not be shown. Only the »No search results in other places« #emptycontent thing. Then there’s also no overlap, and clear feedback.

So we simply hide the list of all contacts? I question if click on a search result will continue to work ...

DeepDiver1975:
Bäääh - one cannot hide a table?

jancborchardt:
Should be possible. display:none or visibility:hidden, on the whole element. Or apply the class `.hidden-visually` to move it off the page.

aspdye:
Any updates?

jancborchardt:
cc @Henni @babelouest @jbtbnl

DeepDiver1975:
TBH: I'd aim for 9.0 having carddav backend in core and a pure js based application - we will add proper search in there then

PR #982, opened 2015-07-13
Fix problem with failing sync of shared contacts

gvde:
syncing shared address books fails (at least to OS X and iOS clients).
Done similar to the changes for calendar commit 3ea6a524fc5c4aa6c3c653795efa408bb18f45dd . Works for me but I really don't know anything about the oc code so someone should carefully review if this is right...


gvde:
I'm all new to github and I have seem to have messed up, mixing up two different patches for different issues. Restarted and now submitted pull request #987 instead which only includes the fix for the failing sync...

Issue #986, opened 2015-07-14
OC 8.2: trailing backslash in name properties


ulfstabe:
When creating a new contact , the name properties are saved in the database (we use mysql) with a trailing backslash. When exported via the remote.php interface this results in a VCard like this:

BEGIN:VCARD
VERSION:3.0
PRODID:-//ownCloud//NONSGML Contacts 0.4.0.0//EN
N:TestContact\;OC\;\;\;;;;;
UID:b467d649-a1e8-4e5d-bc90-36a2eaedfeeb
REV:2015-07-14T17:00:47+00:00
FN:OC TestContact
ORG:ACME
EMAIL;TYPE=HOME:octestcontact@test.de
TEL;TYPE=HOME:0987654321
END:VCARD

rfc2822:
Have the same problem here (fresh OwnCloud 8.1.0, installed for testing purposes)

gvde:
If I create a new contact everything looks fine:

BEGIN:VCARD
VERSION:3.0
PRODID:-//ownCloud//NONSGML Contacts 0.4.0.0//EN
N:TestContact;OC;;;
UID:6a1a3b72-fb81-482e-997f-6fe4a415870c
REV:2015-07-18T18:20:25+00:00
FN:OC TestContact
END:VCARD

But without knowing how you created the contact exactly, with what software/browser etc. it's difficult to tell...  See [contribution guidelines](https://github.com/owncloud/contacts/blob/master/CONTRIBUTING.md)

ainola:
I'm on a Debian 8.1 server with owncloud 8.1. Contacts are randomly getting semicolons added to fields. Here are some example contacts that I just created with the web interface:

```
| 7105 | ainola |       998 | N          | \;aaaaaaaa\;\;\;;;;;                                 |         0 |
| 7106 | ainola |       998 | UID        | df5fb390-a7ab-41dc-a024-7b216fb67dad                 |         0 |
| 7107 | ainola |       998 | FN         | aaaaaaaa                                             |         0 |
| 7108 | ainola |       998 | EMAIL      | asdf@gmail.com                                       |         0 |
| 7109 | ainola |       998 | TEL        | 5555555555                                           |         0 |
| 7110 | ainola |       998 | ADR        | ;;123 fake street;CO;80222;Denver;                   |         0 |
| 7111 | ainola |       999 | UID        | 3a5d397f-07e9-408c-8bc2-96dc144ea070                 |         0 |
[...]

PR #987, opened 2015-07-15
Fix sync issue of shared address books.

gvde:
syncing shared address books fails (at least to OS X and iOS clients).
Done similar to the changes for calendar commit 3ea6a524fc5c4aa6c3c653795efa408bb18f45dd . Works for me but I really don't know anything about the oc code so someone should carefully review if this is right...

### DIFF
--- a/lib/carddav/addressbook.php
+++ b/lib/carddav/addressbook.php
@@ -28,7 +28,8 @@ use OCA\Contacts;
  * This class overrides __construct to get access to $addressBookInfo and
  * $carddavBackend, \Sabre\CardDAV\AddressBook::getACL() to return read/write
  * permissions based on user and shared state and it overrides
- * \Sabre\CardDAV\AddressBook::getChild() and \Sabre\CardDAV\AddressBook::getChildren()
+ * \Sabre\CardDAV\AddressBook::getChild(), \Sabre\CardDAV\AddressBook::getChildren()
+ * and \Sabre\CardDAV\AddressBook::getMultipleChildren()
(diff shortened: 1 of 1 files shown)


gvde:
The initial change was more a dirty workaround than a real fix.
Now I figured what the real problem was: for the initial update OSX uses a multi-get to get the cards quickly. The sabredav multi-get uses getMultipleCards from the Addressbook class which was not overridden by the OCA\Contacts\AddressBook class. Thus it instantiated the sabredav Card class instead which would not allow sharing...

jakobsack:
@Bugahans does this PR solve your WP problems when syncing the contacts?

Bugahans:
I added the function getMultipleChildren to the addresssbook.php but nothing changed for me. I can add a new contact on my wp8 which is synced to owncloud
I can add a contact in the owncloud GUI which is synced to my phone
But a lot of my existing contacts in my owncloud are not shown on my phone.

gvde:
Windows Phone doesn't seem to use multi-get thus this commit is unrelated for the Windows Phone problem. See #998

Garfonso:
I'm pretty positive, that this solved my issue. At least I added this patch (and #986 because I had some other issues with that, too) and now my contacts did sync to my phone.

I had to change a contact in each addressbook to make WP8.1 recognize that there are changes,  though.

mdik:
Hi,

I can confirm that this pull requests fixes things (see https://github.com/owncloud/contacts/issues/1004).

Thank you!

Malte

Issue #985, opened 2015-07-14
Preferred attributes show as "PREF"


gvde:
Seems like the handling of preferred attributes and the parsing of the TYPE=PREF value is incorrect...

### Steps to reproduce
1. Have a contact with no preferred values and 2 e-mail addresses, one "Home" and one "Work". Open it on in your web browser. Edit it.
2. Check the checkbox for before the "Home" e-mail address. 
3. Close the card, i.e. click on "Close".
4. Open the card again.

### Expected behaviour
The home e-mail address should have label "Home" with checkbox set.

### Actual behaviour
The home e-mail address has label "PREF" with checkbox set.

Clearing the checkbox also doesn't work (if you leave the PREF label): removing the checkbox will show the correct label but checkbox still set after reopening...

### Server configuration
**Operating system**: Ubuntu 14.04.2 LTS (GNU/Linux 3.13.0-042stab106.6 x86_64)

**Web server:** Apache 2.4.7-1ubuntu4.4

**Database:** MySQL 5.5.43-0ubuntu0.14.04.1

**PHP version:** php 5.5.9+dfsg-1ubuntu4.11

**ownCloud version:** (see ownCloud admin page) ownCloud 8.1.0 (stable)

**Updated from an older ownCloud or fresh install:** updated from 8.0.4

**List of activated apps:**

Enabled:
 - activity: 2.0.1
 - calendar: 0.7.2
 - contacts: 0.4.0.0
 - files: 1.1.9
 - files_locking: 
 - files_pdfviewer: 0.7
 - files_sharing: 0.6.2
 - files_texteditor: 0.4
 - files_trashbin: 0.6.3
 - files_versions: 1.0.6
 - files_videoviewer: 0.1.3
 - firstrunwizard: 1.1
 - gallery: 0.6.0
 - provisioning_api: 0.2
 - templateeditor: 0.1

Disabled:
 - encryption
 - external
 - files_external
 - user_external
 - user_ldap
 - user_webdavauth

**The content of config/config.php:**

<?php
$CONFIG = array (
  'instanceid' => 'ockoxl25xpg2',
  'passwordsalt' => 'something',
  'secret' => 'something',
  'trusted_domains' => 
  array (
    0 => 'www.example.org',
  ),
  'datadirectory' => '/var/www/owncloud/data',
  'overwrite.cli.url' => 'https://www.example.org/owncloud',
  'dbtype' => 'mysql',
  'version' => '8.1.0.8',
[...]

PR #988, opened 2015-07-15
Fix preferred contact property

gvde:
Bug fixes.

Correctly show property label instead of "PREF" when TYPE=PREF is set and don't forget about the set label on properties after reloading...


LEDfan:
Thanks for your patch, this solves the problem. Fixs #985

PR #992, opened 2015-07-17
back port of pull request owncloud/contacts#987 to stable8.1

gvde:
Fix sync issue of shared address books.


gvde:
Updated for the real fix. See #987

PR #999, opened 2015-07-22
make sure N property is set as multi-valued...

gvde:
Split N property text if it comes as structured text instead of an array...


dynek:
For what it's worth: I just started using the contacts app to sync with my iDevices and saw this issue. Applied the fix and it now works. Thank you.

ghost:
This issue persists when importing Gmail contacts. There is an additional backslash between the first name and the last name. See #1028

LEDfan:
Seems okay to me. Thanks!

PR #1001, opened 2015-07-23
insert correct userid (aka id of owner) into index properties.

gvde:
The index table (oc_contacts_cards_properties) contains the userid of the owner of an entry, which is the owner of the addressbook of the card from which the index entry is taken from.
The updateIndex function, however, always puts in the user id of the current user. This is incorrect when an addressbook is shared allowing edits for others. Whenever a user changes something in a shared card the index entries are rewritten. This means the userid of an index table entry is in fact the user id of the last user editing the entry.
This commit extends updateIndex with a third argument which allows to pass the owner for that index entry which is used when a contact is added or updated.


LEDfan:
Looks good, thanks!

PR #1005, opened 2015-07-30
Fix birthday dtstamp

gvde:
If there is no DTSTAMP set in a calendar object it will automatically set to the current time. Due to this, birthday events are considered changed each time the client checks them.
With this change, DTSTAMP is set the time of last modification of the contact, thus clients won't think the birthday event has been changed unless the contact has been changed...

### DIFF
--- a/lib/contact.php
+++ b/lib/contact.php
@@ -779,6 +779,9 @@ class Contact extends VObject\VCard implements IPIMObject {
 			);
 			$vEvent->DTSTART['VALUE'] = 'date';
 			$vEvent->add('DURATION', 'P1D');
+			$lm = new \DateTime('@' . $this->lastModified());
+			$lm->setTimeZone(new \DateTimeZone('UTC'));
+			$vEvent->DTSTAMP->setDateTime($lm);
 			$vEvent->{'UID'} = $this->UID;
 			$vEvent->{'RRULE'} = 'FREQ=YEARLY';
 			$vEvent->{'SUMMARY'} = $title . ' (' . $date->format('Y') . ')';


gvde:
The easiest way to test whether this patch is working or not is to use curl to retrieve the etags of the birthday entries:

```
$ curl -u myuser -X PROPFIND https://www.example.com/owncloud/remote.php/caldav/calendars/myuser/contact_birthdays/
...
  <d:response>
    <d:href>/owncloud/remote.php/caldav/calendars/myuser/contact_birthdays/shared::1::720.ics</d:href>
    <d:propstat>
      <d:prop>
        <d:getlastmodified>Sat, 11 Jul 2015 10:57:22 GMT</d:getlastmodified>
        <d:getcontentlength>360</d:getcontentlength>
        <d:resourcetype/>
        <d:getetag>"5213da547f367d619a6d6faa77e79f69"</d:getetag>
        <d:getcontenttype>text/calendar; charset=utf-8</d:getcontenttype>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
  </d:response>
</d:multistatus>
```

You can also use the uri of one entry from above to retrieve only the a single entry:

``````
$ curl -u myuser -X PROPFIND https://www.example.com/owncloud/remote.php/caldav/calendars/myuser/contact_birthdays/shared::1::720.ics
<?xml version="1.0" encoding="utf-8"?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:cs="http://calendarserver.org/ns/">
  <d:response>
    <d:href>/owncloud/remote.php/caldav/calendars/myuser/contact_birthdays/shared::1::720.ics</d:href>
    <d:propstat>
      <d:prop>
        <d:getlastmodified>Sat, 11 Jul 2015 10:57:22 GMT</d:getlastmodified>
        <d:getcontentlength>360</d:getcontentlength>
        <d:resourcetype/>
        <d:getetag>"5213da547f367d619a6d6faa77e79f69"</d:getetag>
        <d:getcontenttype>text/calendar; charset=utf-8</d:getcontenttype>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
  </d:response>
</d:multistatus>```

Without the patch the etag "5213da547f367d619a6d6faa77e79f69" will change with each call. That tells the client that the calendar entry has changed. With the patch the etag remains the same until you make any change to the vcard for that birthday entry. Then you should see a new etag.
``````

georgehrke:
code changes look good,
@gvde please squash the commits

gvde:
@georgehrke I didn't even know I could squash... I hope I did it right. Looks correct to me...

georgehrke:
> I hope I did it right.

Yes, you did.

Thanks for your contribution :)

PR #1041, opened 2015-09-10

gvde:
And only set REV if something has been repaired.
The validate functions has an UPGRADE option which allows 2.1 cards to be upgraded to 3.0. Until now it did not check whether the card to be "upgraded" is in fact 2.1 or maybe 3.0 or even 4.0.
It called decodeProperty and fixPropertyParameters on all properties of that card. While those functions may work good for 2.1 card properties they definitely correctly work for the 4.0 PHOTO uri values. In addition, it doesn't seem a good idea to "downgrade" 4.0 cards to 3.0 as we may loose some information.
With this change validate checks the current version of the card and only if it's 2.1 it will upgrade the card to 3.0...

### DIFF
--- a/lib/vobject/vcard.php
+++ b/lib/vobject/vcard.php
@@ -187,24 +187,7 @@ class VCard extends VObject\Component\VCard {
 	public function validate($options = 0) {
 
 		$warnings = array();
-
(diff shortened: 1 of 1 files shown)


gvde:
Closed in favor of #1054

PR #1042, opened 2015-09-10
Backport of #1041 to stable8.1

gvde:


gvde:
Closed in favor of #1067

Issue #1027, opened 2015-08-19
Shorten Display of Birthdays in Calendar


deMattin:
The birthdays shown in Calendar are very long and in some languages don't begin with the name. So depending on the displayable space, in month or week view (on smartphone) there's only shown "Geburtstag v...." in german for example and you have to go to detail view to see, who has a birthday.

What do you think about replacing the "Geburtstag von {name}" or "{name}'s Birthday" with the UTF8 code for a "birthday cake" (1F382) following by name and year?
I've seen tis in other apps, too and it's very intuitive, I think.

I tested the look by replacing the translation in the l10n files for my language (german) with the cake-code string: "&#x1f382; {name}" and it displays and looks very fine.
But this way it only works for displaying in owncloud web app.
The most issue is with the birthday in a caldav synced calendar where `{name}'s Birthday` or in german the `Geburtstag von {name}` is shown and often cut off disadvantageous in some display scenarios.

__So there is a be a better place to replace this in source for ALL languages with UTF8 birthday cake symbol as part of the entry.
But it only makes sense to investigate further, if core team likes this idea.__

Btw. This is an issue with contacts and not calendar, because contacts generates these calendar entries in owncloud!

deMattin:
My solution, that should work for any language with no issues in log and working.
I need also to update language files because I'm no coder and so don't know, how to rewrite the code in config.php, not to search the language files for translation.
So without also editing language files you get log entries like "translation not found".

For coder from team this would be a 2 line edit in few seconds in config.php also with escaped UTF8 character, but there seems to be no interest and so I do it with my skills and post it here, if someone other likes to change this, too ... ;)

Add in ``apps/contacts/l10n/`` somewhere a new translation:
```
    "{name}" : "{name}",
```
in both language files for your language (Germany: `de.js` and `de.json` - I don't know where the german `de_DE` files are used for ... not for calendar)

in `apps/contacts/lib/contact.php`
change line 766 from
```
				App::$l10n->t('{name}\'s Birthday')
```
to
```
				App::$l10n->t('{name}')
```
and line 784 from
```
			$vEvent->{'SUMMARY'} = $title . ' (' . $date->format('Y') . ')';
```
to
```
			$vEvent->{'SUMMARY'} = 'ðŸŽ‚ ' . $title . ' (' . $date->format('Y') . ')';
```
[...]

PR #1045, opened 2015-09-14
Shorten display of contact's birthdays in Calendar

deMattin:
Replaces "birthday" with a UTF8 birthday cake;
works language independend;
also synchronizes well to PC and smartphone/tablet
https://github.com/owncloud/contacts/issues/1027

### DIFF
--- a/lib/contact.php
+++ b/lib/contact.php
@@ -791,7 +791,7 @@ class Contact extends VObject\VCard implements IPIMObject {
 		if ((string)$birthday) {
 			$title = str_replace('{name}',
 				strtr((string)$this->FN, array('\,' => ',', '\;' => ';')),
-				App::$l10n->t('{name}\'s Birthday')
+                		'{name}'
 			);
 			try {
 				$date = new \DateTime($birthday);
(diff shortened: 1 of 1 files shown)


deMattin:
If integration is accepted by core team, it can also simply be backported to stable 8.1.
(I would make a pull request then, if desired)

jancborchardt:
@babelouest @LEDfan @jbtbnl @schiesbn please review! :)

DeepDiver1975:
Is this 4 byte string stored in the db? If yes we cannot do this - mysql only allows 3byte chars (the way we use it)

DeepDiver1975:
bildschirmfoto von 2015-10-12 14-01-18

bildschirmfoto von 2015-10-12 14-01-36

DeepDiver1975:
I'm honestly not really impressed by this

jakobsack:
I agree with Thomas, I don't like the cake symbol either.

If it is necessary to shorten the birthday entries I'd rather use the star (*) which is not only a ASCII sign but also widely recognized as a symbol for the date of birth. It could look like `Jakob Sack (*1986)`. But that's just my point of view.

deMattin:
Replacing the cake with a star at the year is ok, also.
Everything cutting off "Geburtstag von ..." is better than it is actually!
Many people like to shorten this as you can see in the issue discussion..
@DeepDiver1975 please rethink about shorting this.and implement it in a way you like.

DeepDiver1975:

PR #1046, opened 2015-09-15
Use uppercase DATE for property parameter VALUE

gvde:
as some clients expect uppercase.
Also use DTEND instead of DURATION as that seems how most client generate one-day events.


jakobsack:
:+1: Code looks fine, but I don't understand the issue travis has. 
@DeepDiver1975 @schiesbn @rullzer @PVince81 please review.

jakobsack:
Works as expected when syncing with Windows 10

PR #1048, opened 2015-09-15
Fix #1043; Importing empty BDAY vcard entry evaluates to current date.

### LINKED ISSUE #1043: Import vcard if empty BDAY it gets value of actual date [v.0.4.0.0]
On import my contacts from vcard v.3.0 (about 165 from one file) every contact that got no entry at "BDAY;VALUE=DATE:" is getting birthday of current date o_O.
I tried to use vcard standard and vcard google import option with the same result.

I would expect an empty bday value instead of current date.

OwnCloud 8.1.1 (stable), Contacts v0.4.0.0

ChrisPHL:


lib/contact.php
```diff
@@ -577,6 +577,11 @@ public function setPropertyByName($name, $value, $parameters=array()) {
 		// TODO: parameters are ignored for now.
 		switch ($name) {
 			case 'BDAY':
+				if (empty($value)) {
+					// We don't want "New \DateTime($value)" evaluate to current date because of empty value.
+					// So returning false at this point.
+					return false;
```
gvde, line 583: That's a very bad idea and will cause an Internal Server error in contactcontroller.php line 139.

gvde:
As I wrote in #1043: instead of implementing workaround for broken vcards which have been converted before from a broken Android export it would be better to think about a workaround to get the broken export imported.

DeepDiver1975:

PR #1057, opened 2015-09-28
Import legacy share dialog code

PVince81:
This will make the share dialog work as before

@DeepDiver1975 @schiesbn @rullzer


PVince81:
Note: this only brings it back, it doesn't fix the previously existing layout issues...

rullzer:
Seems to be working here.
:+1:

PVince81:
@rullzer thanks

Issue #1007, opened 2015-07-30
Contact Pictures doesn't sync from Android to OC: "OC_Image->loadFromBase64, could not load"


fabrice-regnier:
Hi folks,

I hope i'm in the right place. I won't be upset if you throw me out ;)

OC server: 8.1.0-3
Contacts: 0.4.0.0

Client : Contact 
Davdroid: 0.8.1
Android: 5.02 and 4.2.2

If i put a picture for a contact on the oc server (https://my_ip/index.php/apps/contacts/) then the sync towards the android client works ok.
If i put a picture on the android contact client, i can see the picture some seconds then it disappears and lets the Letter come back. I guess the sync towards the server didn't work.

In the log (/var/www/owncloud/data/owncloud.log), i can read:

{"reqId":"JorRGcvISWTcV0BVwk9\/","remoteAddr":"80.12.55.138","app":"core","message":"OC_Image->loadFromBase64, could not load","level":0,"time":"2015-07-30T15:26:13+00:00","method":"PUT","url":"\/remote.php\/carddav\/addressbooks\/fabrice.regnier\/contacts\/c75081d5-cbb6-41b1-8c73-7e8999e32170.vcf"}

If you want more logs, let me know,

regards,

f.

by the way, thanks a lot for this great app ;)

gvde:
Check how the content of the vcard looks like. Could be the same problem as #994

fabrice-regnier:
Yes, i guess it's the same issue. Here is the vcard content downloaded from oc, after i sync from android to oc. The photo seems to be there but broken.

BEGIN:VCARD
VERSION:3.0
PRODID:DAVdroid/0.8.1 (ez-vcard/0.9.6)
UID:c75081d5-cbb6-41b1-8c73-7e8999e32170
FN:Nat
N:;Nat;;;
CATEGORIES:Famille
PHOTO:dataimage/jpegbase64/9j/4AAQSkZJRgABAQAAAQABAAD/2wBDAAgGBgcGBQgHBwcJC
 QgKDBQNDAsLDBkSEw8UHRofHh0aHBwgJC4nICIsIxwcKDcpLDAxNDQ0Hyc5PTgyPC4zNDL/2wB
[zap]
REV:2015-07-30T16:00:26+00:00
END:VCARD

I guess i should close this ticket ?

davux:
I'm having the same issue:
- Owncloud 8.1.0-2 according to Ubuntu package, 8.2 according to help.
[...]

Issue #994, opened 2015-07-18
Contact pictures have questionnable PHOTO tag when retrieved


robjk:
Hi,

I think this is a bug that lies on the owncloud/contacts side of things but I could be wrong.

I'm using emClient 6.0.22344.0
owncloud/core version 8.1.0 I believe (downloaded from releases today)
contacts version 8.1 (also downloaded today)

Steps to reproduce (as followed in the logfile below)
1) Start with no contacts defined in owncloud or in emClient
2) Create a new contact in emClient named "Test"
3) Refresh and confirm contact appears in owncloud/contacts
4) Add an image to the test contact in emClient

The result of step 4 is that the contact disappears from emClient, and the avatar in contacts becomes:
![capture](https://cloud.githubusercontent.com/assets/5565386/8763277/53e83f48-2d8c-11e5-8511-7ed224beb7f6.PNG)
(where you'd normally see a large "T" it's just a white background)

Looking at the CardDav logs I can see that when emClient submits the .vcf, the PHOTO tag appears as such:
`20:23:09.617|001|   Synchronizer.Modify:  avatar class updated
20:23:09.627|014|   C: PutCommand http://192.168.1.124/owncloud/remote.php/carddav/addressbooks/rob/contacts/ea875ad9-5451-42a0-b0e3-703bae86f434.vcf
20:23:09.627|014|   BEGIN:VCARD
20:23:09.627|014|   VERSION:4.0
20:23:09.627|014|   UID:urn:uuid:a8536c6f-10cf-436a-97d7-9cdd136cfa7d
20:23:09.627|014|   N;SORT-AS="Contact",Test:Contact;Test;;
20:23:09.627|014|   FN:Test Contact
20:23:09.627|014|   REV:2015-07-18T19:22:43+00:00
20:23:09.627|014|   PHOTO:data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAaQAAAGkCAYAAAB+TFE1AAAA
`

but the response from the server appears to strip out the punctuation (colons/semi-colon/comma) between "data", "image/png", "base64" and the base64 encoded payload:
`20:23:10.006|014|   C: GetCommand http://192.168.1.124/owncloud/remote.php/carddav/addressbooks/rob/contacts/ea875ad9-5451-42a0-b0e3-703bae86f434.vcf
[...]

PR #1067, opened 2015-10-15
Fix multi version and only upgrade 2.1 cards.

gvde:
Backport of #1054 
Make sure to replace PRODID and VERSION instead of adding
additional ones...

If multiple VERSION values are set, remove all and set only the value
of the first.
If card has been repaired change PRODID to own appinfo.

satisfy Scrutinizer

### DIFF
--- a/lib/connector/importvcardconnector.php
+++ b/lib/connector/importvcardconnector.php
@@ -143,7 +143,13 @@ class ImportVCardConnector extends ImportConnector{
 				}
 			} else {
 				$property = clone $sourceProperty;
-				$dest->add($property);
+				// VERSION and PRODID are set by default in any new empty vcard object.
+				// Thus it needs to be replaced instead of simply added as additional value.
+				if ( $property->name === 'PRODID' || $property->name === 'VERSION' ) {
+					$dest->__set($property->name, $property);
+				} else {
+					$dest->add($property);
+				}
 			}
 		}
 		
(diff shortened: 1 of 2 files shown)


h-2:
Yes, this fixes my problems, please merge it :+1:

ColinFinck:
Just upgraded from ownCloud 8.1.3 to 8.2.0 in the hope to fix issue #1007 (aka #994), but was shortly disappointed afterwards when noticing that my Contact photos were still being deleted. I'm successfully running an ownCloud installation with the Contacts app since 6.x and the aforementioned bug destroys most of the great experience.

Manually applying the patch provided here solves the problem.
Please merge this back and backport the fix as soon as possible to all supported versions and provide new stable releases.
It would be a disaster if even more people running the latest stable ownCloud version lose parts of their Contact database due to this nasty bug.

KekcuHa:
I have applyed provided patch on OC 8.2.2 and it solved the problem of the sync android contacts with photos.
gvde, thank you very much!

h-2:
What is preventing this from being merged? Isn't like everyone affected?

gvde:
There is no maintainer of this app. Thus is takes a while until someone else does the merge.

fossxplorer:
I can do the merge since others have tested the PR. Is it OK?
What about backport to stable8.2?

tribut:
:+1:

gvde:
master PR is #1054 
stable8.2 PR is #1068

kallegit1950:
I have tested #1068 on my OC installation and it fixes an issue that a sync via DavDroid removes attached photos.

fossxplorer:
@gvde Travis CI fails on https://github.com/owncloud/contacts/pull/1054, but https://github.com/owncloud/contacts/pull/1068 is now merged.

gvde:
@fossxplorer just rerun travis for #1054 and it passes now... (It failed in October, because core was on 9.0 but contacts app master was not compatible set to be compatible with 9.0...)

fossxplorer:
@gvde 1054 merged. Thx a lot for your help!

PR #1068, opened 2015-10-15
Fix multi version and only upgrade 2.1 cards.

gvde:
Backport of #1054

Make sure to replace PRODID and VERSION instead of adding
additional ones...

If multiple VERSION values are set, remove all and set only the value
of the first.
If card has been repaired change PRODID to own appinfo.

satisfy Scrutinizer

### DIFF
--- a/lib/connector/importvcardconnector.php
+++ b/lib/connector/importvcardconnector.php
@@ -143,7 +143,13 @@ class ImportVCardConnector extends ImportConnector{
 				}
 			} else {
 				$property = clone $sourceProperty;
-				$dest->add($property);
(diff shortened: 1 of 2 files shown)


kallegit1950:
if it helps, I have tested this fix on 8.2. and it helps fixing a sync issue with davdroid that photos are deleted. Can this PR be commitet that everyone benefits from it. Is there anything else I can do to test and expedite this issue?

grunsch:
I have tested this fix on 8.2.2 and it helps fixing a sync issue with davdroid that photos are deleted.

pboesch:
The fix on 8.2.2 works too for me. Thanks

PR #1076, opened 2015-10-21
Upgrade max-version to 9.0 in appinfo

Henni:


DeepDiver1975:
:+1:

DeepDiver1975:
We also need to adjust travis.yml which tries to test against 8.1

Henni:
BTW: contacts currently doesn't work on 9.0 as the dav code in core changed. So maybe we should wait with updating the info.xml

DeepDiver1975:
:-1: the app will not work in oc 9.0

Issue #1110, opened 2015-12-01
Wrong Birthyear in calendar


awrede:
If a contact's birthday is 31.Dec then the calendar entry generated will show the wrong birth year in the title. It' is caused by the year of the DTEND record being used to, with is the birthday +1D.  Trivial fix attached.

[patch0.txt](https://github.com/owncloud/contacts/files/48725/patch0.txt)

gvde:
Created #1111 for master branch.

PR #1111, opened 2015-12-01
Show correct year of birth when birthday is Dec. 31st.

gvde:
Fix from/for #1110


DeepDiver1975:

PR #1115, opened 2015-12-18
use public API method instead of deprecated private namespace method

MorrisJobke:
cc @DeepDiver1975 @LukasReschke


LukasReschke:
:+1:

